PR #5, opened 2020-05-20
Add UseCase for load newer messages

maxmaxandr:
I've added UseCase for load newer messages as it implemented for load older messages. I need it for the ability to start to chat not only from the end of the list but from any position in the MessageListView.


tschellenbach:
Makes a lot of sense +1

PR #11, opened 2020-06-16
Bug / NotificationMarkReadEvent crash 

samiuelson:
### Reason of crash:
Handling NotificationMarkReadEvent was causing a crash because `event.user` value was being sent null.

### Solution: 
Using the current user which was set to the ChatDomain instead of relying on event's data.

### DIFF
--- a/livedata/build.gradle
+++ b/livedata/build.gradle
@@ -3,8 +3,8 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
 apply plugin: 'org.jetbrains.dokka'
-apply plugin: 'com.github.dcendents.android-maven'
 apply plugin: 'jacoco-android'
+apply plugin: 'maven-publish'
 
 //https://github.com/arturdm/jacoco-android-gradle-plugin
 
@@ -33,7 +33,7 @@ android {
         minSdkVersion 21
         targetSdkVersion 28
         versionCode 1
-        versionName "0.6.4"
+        versionName "0.6.5"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
     compileOptions {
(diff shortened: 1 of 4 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -791,10 +791,16 @@ class ChannelControllerImpl(
...
+            is MessageReadEvent -> {
                 val read = ChannelUserRead(event.user!!, event.createdAt!!)
                 updateRead(read)
             }
+            is NotificationMarkReadEvent -> {
+                val user = domainImpl.currentUser
+                val date = event.createdAt!!
+                val read = ChannelUserRead(user, date)
+                updateRead(read)
+            }
```
tschellenbach, line 803: looks good

livedata/build.gradle
```diff
@@ -124,7 +124,23 @@ dependencies {
     androidTestImplementation "org.mockito:mockito-core:3.2.4"
     androidTestImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
     androidTestImplementation 'org.amshove.kluent:kluent:1.61'
-
-
 }
 
+afterEvaluate {
```
tschellenbach, line 129: this seems awesome, how does it work? got a good blogpost to read for me?

samiuelson: hmm, here is some info about mavenLocal:
https://proandroiddev.com/tip-work-with-third-party-projects-locally-with-gradle-961d6c9efb02
android-related customizations:
https://developer.android.com/studio/build/maven-publish-plugin

Consuming mavenLocal dependency in the target project is the same as any other else, e.g.
``` groovy
dependencies {
    api 'com.github.GetStream:stream-chat-android-livedata:0.6.5-SNAPSHOT'
}
```

PR #12, opened 2020-06-29
Filter messages that belong to a thread

JcMinarro:
Filter messages that belong to a thread and are not marked as `showInChannel`. They shouldn't be shown on the channel itself


tschellenbach:
👍

PR #14, opened 2020-07-10
Force emptyList when data is not present.

emitenma:
Make sure we return emptyList when we don't have data.
emptyList is inmutable vs empty arrayList() which is mutable.

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
@@ -203,9 +203,10 @@ class ChannelControllerImpl(
 
     fun sortedMessages(): List<Message> {
         // sorted ascending order, so the oldest messages are at the beginning of the list
-        val messageMap = _messages.value ?: mutableMapOf()
-        val messages = messageMap.values.sortedBy { it.createdAt }.filter { hideMessagesBefore == null || it.createdAt!! > hideMessagesBefore }
-        return messages
+        _messages.value?.let {
+            return it.values.sortedBy { it.createdAt }.filter { hideMessagesBefore == null || it.createdAt!! > hideMessagesBefore }
+        }
+        return emptyList()
     }
 
     fun removeMessagesBefore(t: Date) {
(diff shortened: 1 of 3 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -203,9 +203,10 @@ class ChannelControllerImpl(
 
     fun sortedMessages(): List<Message> {
         // sorted ascending order, so the oldest messages are at the beginning of the list
-        val messageMap = _messages.value ?: mutableMapOf()
-        val messages = messageMap.values.sortedBy { it.createdAt }.filter { hideMessagesBefore == null || it.createdAt!! > hideMessagesBefore }
-        return messages
+        _messages.value?.let {
+            return it.values.sortedBy { it.createdAt }.filter { hideMessagesBefore == null || it.createdAt!! > hideMessagesBefore }
+        }
+        return emptyList()
```
JcMinarro, line 209: What do you think about to have only one return?
```suggestion
        return _messages.value?.let {
            it.values.sortedBy { it.createdAt }.filter { hideMessagesBefore == null || it.createdAt!! > hideMessagesBefore }
        }
        ?: emptyList()
```

livedata/src/test/java/io/getstream/chat/android/livedata/controller/ChannelControllerImplTest.kt
```diff
@@ -108,4 +115,26 @@ class ChannelControllerImplTest : BaseDomainTest() {
             ) was called
         }
     }
+
+    @Test
+    fun `Should return an empty channel when data is not present`() {
```
JcMinarro, line 120: Good to see more tests in our code ;)

tschellenbach:
The UI is typically different for data is not present yet, vs an empty list. After this change how do you know that no data is present?

emitenma:
Hi Thierry,
the change was to align inconsistent management of the empty arrays.
The UI shouldn't be affected, as in both cases, we return an empty lists.
The difference between them consist on mutability.

PR #15, opened 2020-07-14
Fix NotificationAddedToChannelEvent event handling

samiuelson:


samiuelson:
> probably also good to update the test data in test data helper to reflect the right event structure

Yes, I'm working on tests now.

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -380,7 +381,7 @@ class ChatDomainImpl private constructor(
 
     fun channel(cid: String): ChannelControllerImpl {
         val parts = cid.split(":")
-        check(parts.size == 2) { "Received invalid cid, expected format messaging:123, got $cid" }
+        check(parts.size >= 2) { "Received invalid cid, expected format messaging:123, got $cid" }
```
JcMinarro, line 384: Why do we accept more than 2 parts on the `cid`? 
Is there any case where `cid` is something different than `xxx:xxx`

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -467,28 +468,12 @@ class ChatDomainImpl private constructor(
...
-            val events = response.data()
-            allEvents.addAll(events)
-            if (events.size <= limit || offset >= maxEvents) {
-                break
-            }
-            offset += limit
+    private fun queryEvents(cids: List<String>): List<ChatEvent> {
+        val response = client.getSyncHistory(cids, syncState?.lastSyncedAt ?: NEVER).execute()
+        if (response.isError) {
+            throw response.error()
```
JcMinarro, line 474: Instead of throw an exception, is not possible to return an error?

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -672,7 +672,9 @@ class ChannelControllerImpl(
             }
             is MemberAddedEvent, is MemberUpdatedEvent, is NotificationAddedToChannelEvent -> {
                 // add /remove the members etc
-                upsertMember(event.member!!)
+                event.channel?.members?.let {
```
JcMinarro, line 675: Good to see NPE-Safe code ;)

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -803,10 +805,18 @@ class ChannelControllerImpl(
...
+        _members.postValue(channelMembers)
+    }
+
     fun upsertMember(member: Member) {
-        val copy = _members.value ?: mutableMapOf()
-        copy[member.user.id] = member
-        _members.postValue(copy)
+        val channelMembers = _members.value ?: mutableMapOf()
+        channelMembers[member.user.id] = member
+        _members.postValue(channelMembers)
```
JcMinarro, line 819: What do you thing to reuse the previous function?
Then we have the "matching logic" only one place.
You could dome something like:

```suggestion
        upsertMembers(listOf(member))
```

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
tschellenbach: dont we need pagination on this? depends on how many items we can fetch at once of course. good to check with the backend team.

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
samiuelson: Channel name regex doesn't allow ":" in channel name now. So, I'm switching back to "==".

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
samiuelson: The current implementation of sync endpoint client in LLC doesn't support any pagination. I'll check it with the backend team when working on the sync itself.

PR #16, opened 2020-07-15
Bugfix/cas 161 cid mutability causes crashes

emitenma:
Access the CID of a message, ALWAYS with Channel as a backup.

We should have inmutability for CID in general, and this is the first step to unify the access to its value.

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
@@ -1,4 +1,4 @@
-package io.getstream.chat.android.livedata
+package io.getstream.chat.android.livedata.extensions
 
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.errors.ChatNetworkError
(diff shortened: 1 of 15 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -28,16 +28,13 @@ fun ChatEvent.getCid(): String? {
 /**
  * cid is sometimes devent as message.cid other times as message.channel.cid
  */
-fun Message.getCid(): String? {
-
-    var cid = this.cid
-
-    if (cid.isNullOrEmpty()) {
-        cid = this.channel?.cid
+fun Message.getCid(): String =
```
JcMinarro, line 31: What do you think about to override the getter of `Message#cid` to do this logic and be sure it is always used on the proper way?

It could be implemented like:
```
data class Message(private val _cid: String) {
  var cid: String = _cid
    get() = field.takeUnless { it.isEmpty() } ?: channel.cid
}
```

emitenma: Mmm the problem with this, is taht we would be adding logic to a Data Class, which shouldn't have logic at all.
I usually prefer this kind of things as extended logic outside the data class (like the open-closed principle). 

Still, at this moment, we could do both. I don't mind changing, if we keep consistency.

tschellenbach:
Trying to remember here, but i think there is also the issue where a message object received on an event sometimes doesn't have the CID (but the event does)

PR #20, opened 2020-07-24
[CAS-174] Background data sync

samiuelson:


livedata/build.gradle
```diff
@@ -138,7 +138,7 @@ afterEvaluate {
                 // Custom attributes of the publication
                 groupId = project.group
                 artifactId = 'stream-chat-android-livedata'
-                version = '0.6.8-SNAPSHOT'
+                version = '0.6.12-SNAPSHOT'
```
JcMinarro, line 141: Can we come back to the previous values as we didn't increase the version yet?

livedata/src/main/java/io/getstream/chat/android/livedata/ChatNotificationConfigOffline.kt
```diff
@@ -1,45 +1,33 @@
...
         val channelId = data.get("channel_id").toString()
         val channelType = data.get("channel_type").toString()
         val cid = "$channelType:$channelId"
 
-        logger.logI("onFirebaseMessage started a worker to update channel $cid")
-        //  Start background work manager
-        val constraints = Constraints.Builder()
-                .setRequiredNetworkType(NetworkType.CONNECTED)
-                .build()
+        Intent(context, SyncService::class.java).apply {
```
JcMinarro, line 22: What do you think about to move "Intent creation" into `SyncService`?
Following this approach we can update the needed attributes to be instanciated and be sure all place where it is used add them.

You can create an static method into this class like:
```
fun createIntent(context: Context, cid: String): Intent = Intent(context, Service::class.java).apply { putExtra(EXTRA_CID, cid) }
```

It allows us to not need to share the "attribute name" on different parts of our code, and the service knows exactly which attributes expects

livedata/src/main/java/io/getstream/chat/android/livedata/service/SyncService.kt
```diff
@@ -0,0 +1,64 @@
...
+            .apply {
+                startForeground(NOTIFICATION_ID, this)
+            }
+
+        try {
+            intent?.getStringExtra(EXTRA_CID)?.let {
+                // TODO: ChatDomain is null when app gets killed.
+                // That's why it's wrapped in try clause atm.
+                // Need to store api key to re-create ChatClient + ChatDomain here.
+                ChatDomain.instance().apply {
```
JcMinarro, line 31: All this logic (For now very simple, but it will do some extra job) can be extracted into another collaborator, then we are able to test our logic without needing to instantiate our service

samiuelson: Good idea, I'll extract it to a new use case perhaps, in next PR

PR #22, opened 2020-07-29
Local QuerySort

### LINKED ISSUE #21: Incorrect work of QuerySort
SDK 4.2.10-beta

I just created a screen with channels and use the next code:
```
    val filter = Filters.and(Filters.`in`("members", userId))
    val querySort = QuerySort().desc("last_message_at")
    channelsViewModel?.setQuery(filter, querySort)
```
As a result, I got all necessary channels but with unexpected sorting.

JcMinarro:
`QuerySort` is an abailable parameter we use to sort the result query as the user expect, but It was only implemented for "remote queries", returning wrong result when some local data was stored on the device.
Because our queries are paginated, this paremeter needs to be added to our pagination class to apply the pagination on a proper way.
`QuerySort` was thought to be used on an API Request, where the name of the field is enough to perform the query. In local we have some problems, becasue this field is indicated as an String that indicate the name of the field where the sort should be done. To achieve this goal, I needed to use reflection ops over our model data to be able to obtain the proper fields and sort by them.
Because this name is an String, it could be whatever value, and we need to offer a default value where the field name doesn't exist on our model

Fix #21
Fix GetStream/stream-chat-android#582

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -337,8 +337,7 @@ class ChatDomainImpl private constructor(
     /** stores the mapping from cid to channelRepository */
     var activeChannelMapImpl: ConcurrentHashMap<String, ChannelControllerImpl> = ConcurrentHashMap()
 
-    /** stores the mapping from cid to channelRepository */
-    var activeQueryMapImpl: ConcurrentHashMap<QueryChannelsEntity, QueryChannelsControllerImpl> = ConcurrentHashMap()
+    private val activeQueryMapImpl: ConcurrentHashMap<String, QueryChannelsControllerImpl> = ConcurrentHashMap()
 
     fun isActiveChannel(cid: String): Boolean {
         return activeChannelMapImpl.containsKey(cid)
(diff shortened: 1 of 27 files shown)


tschellenbach:
This is not the right approach.

1. You can't fully emulate the query behaviour locally:

A. It queries data that we don't have locally
B. Fully supporting this Mongo style syntax is complex
C. Query behaviour can change server side, we don't want to maintain this in 2 places.

2. The idea here is to rely on the list of channels that the server returns when querying

tschellenbach:
I spoke to soon, the above applies to filtering, but there might be a good use case for implementing sorting locally.

* The channel data changes while the app is open, we need to resort as data changes. So sorting locally is preferred. 
* (This mostly doesn't apply to filtering, very rare for apps to have a flow where the channel changes in a way that it no longer matches the filter)

So thinking about implementing sorting locally it's good to understand what the API supports. This is from our docs:

```
The sorting uses for the channels matching the filters. sorting is based on field and direction, multiple sorting options can be provided. You can sort based on last_updated, last_message_at, updated_at, created_at, member_count, unread_count or has_unread(unread status). Direction can be ascending (1) or descending (-1)
```

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -449,23 +450,16 @@ class ChatDomainImpl private constructor(
...
-                    queryChannelsEntity,
-                    queryChannelsEntity.filter,
-                    queryChannelsEntity.sort,
-                    client,
-                    this
-                )
-            activeQueryMapImpl[queryChannelsEntity] = queryRepo
+        sort: QuerySort
+    ): QueryChannelsControllerImpl =
+        activeQueryMapImpl.getOrPut("${filter.hashCode()}-${sort.hashCode()}") {
```
tschellenbach, line 455: see query.id 
https://github.com/GetStream/stream-chat-android-livedata/blob/master/livedata/src/main/java/io/getstream/chat/android/livedata/entity/QueryChannelsEntity.kt#L17

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
```diff
@@ -17,11 +17,11 @@ interface QueryChannelsController {
     /**
      * The filter used for this query
      */
-    var filter: FilterObject
+    val filter: FilterObject
```
tschellenbach, line 20: 👍

livedata/src/test/java/io/getstream/chat/android/livedata/repository/QueryChannelsImplRepositoryTest.kt
```diff
@@ -17,7 +17,7 @@ class QueryChannelsImplRepositoryTest : BaseDomainTest() {
     @Test
     fun testInsertAndRead() = runBlocking(Dispatchers.IO) {
         val queryChannelsEntity = QueryChannelsEntity(data.filter1, QuerySort())
-        queryChannelsEntity.channelCids = sortedSetOf("a", "b", "c")
+        queryChannelsEntity.channelCids = listOf("a", "b", "c")
```
tschellenbach, line 20: be sure to test in the sample app that we don't get the same channel showing up multiple times...

JcMinarro: It is handled when new channels are added

PR #23, opened 2020-07-30
Handle error when a channel is created with a bad CID

JcMinarro:
When a Channel was created, we didn't handle an invalid CID on the process, and the SDK threw an unchecked exception that could crash an app. It has been updated to handle that exception and return a ResultError on that case.


tschellenbach:
Nice, yeah i think this is a good fix.

PR #24, opened 2020-07-30
CAS-177 Allow to store apiKey when app was killed

samiuelson:
* Encrypted store for apiKey and token
* Instantiate ChatDomain and ChatClient when app was killed


livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -81,12 +83,21 @@ interface ChatDomain {
...
         fun database(db: ChatDatabase): Builder {
             this.database = db
             return this
         }
 
+        fun backgroundSyncEnabled(apiKey: String, userId: String, userToken: String): Builder {
+            this.backgroundSyncConfig = BackgroundSyncConfig(apiKey, userId, userToken)
+            if (apiKey.isEmpty() || userId.isEmpty() || userToken.isEmpty()) {
+                throw IllegalArgumentException("ChatDomain.Builder::backgroundSyncEnabled. apiKey, userId, and userToken must not be empty.")
+            }
```
tschellenbach, line 97: doesnt the client object already expose the apiKey and token, and the user the user id?
just that this leaves rooms for bugs if you pass something different than the client/user objects

tschellenbach: looks great other than that

samiuelson: The client stores those data (apiKey, token) in memory, but doesn't expose it. 
Would it be safe if it did? I agree it would allow avoiding potential troubles and would simplify our API.
cc @JcMinarro 

In UI sdk this data is available in `Chat.Builder` internally. So the UI API gets only a simple additional prop `Builder.backgroundSyncEnabled: Boolean`:
```
Chat.Builder(appConfig.apiKey, this).apply {
    backgroundSyncEnabled = true
     // ... 
}.build()
```

In case of `ChatDomain.Builder` there are no such data available internally.

JcMinarro: I would prefer to have some method like:
```
fun backgroundSyncEnabled(): Builder = apply {
  backgroundSyncEnabled = true
}
```
And try to get all this values later when they are started.
All this data could be stored when we set the current user to our chat entity, just on that moment we can trigger the Store method and so on

samiuelson: Yep, that's what I'm doing now inside `ChatImpl.setUser()`. 
```
@Override
    public void setUser(@NotNull User user,
                        @NotNull String userToken,
                        @NotNull InitConnectionListener callbacks) {
        // ...
        final ChatDomain.Builder domainBuilder = new ChatDomain.Builder(context, client, user);
        if (backgroundSyncEnabled) {
            domainBuilder.backgroundSyncEnabled(apiKey, user.getId(), userToken);
        }
        domainBuilder.build()
        
        client.setUser(user, userToken)
       // ...
    }
```
But on the `ChatDomain.Builder` side, i need to pass those data explicitly here: `domainBuilder.backgroundSyncEnabled(apiKey, user.getId(), userToken)`

Other option would be to expose `apiKey` and `userToken` values by ChatClient and obtain those internally in ChatDomainImpl.

JcMinarro: Because you are doing it into the method `setUser()`, you already have the token itselft, you don't need to add it as a new attribute. right?
On the other hand, it is an implementation details of our `ChatDomain` instance, and shouldn't be accesible by devs, we should reduce the visibility of this method

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -192,6 +205,18 @@ class ChatDomainImpl private constructor(
         initClean()
     }
 
+    private fun storeBackgroundSyncConfig(
+        backgroundSyncConfig: BackgroundSyncConfig,
+        context: Context
+    ) {
+        if (BackgroundSyncConfig.UNAVAILABLE != backgroundSyncConfig) {
+            EncryptedBackgroundSyncConfigStore(context.applicationContext).run {
```
JcMinarro, line 213: Could we inject this dependency instead of creating it here?
It will help us on test side

livedata/src/main/java/io/getstream/chat/android/livedata/service/sync/BackgroundSyncConfigStore.kt
```diff
@@ -0,0 +1,59 @@
+package io.getstream.chat.android.livedata.service.sync
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.security.crypto.EncryptedSharedPreferences
+import androidx.security.crypto.MasterKey
+
+interface BackgroundSyncConfigStore {
```
JcMinarro, line 8: Do we really need an interface here?

livedata/src/main/java/io/getstream/chat/android/livedata/service/sync/BackgroundSyncConfigStore.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun put(config: BackgroundSyncConfig)
+    fun get(): BackgroundSyncConfig
+    fun clear()
+}
+
+class EncryptedBackgroundSyncConfigStore(context: Context) : BackgroundSyncConfigStore {
```
JcMinarro, line 14: This is an internal class that shouldn't be used by other devs, we should keep them `internal`

livedata/src/main/java/io/getstream/chat/android/livedata/service/sync/BackgroundSyncConfigStore.kt
```diff
@@ -0,0 +1,59 @@
...
+            putString(KEY_USER_TOKEN, config.userToken)
+        }.apply()
+    }
+
+    override fun get(): BackgroundSyncConfig = prefs.run {
+        val apiKey = getString(KEY_API_KEY, "")
+        val userId = getString(KEY_USER_ID, "")
+        val userToken = getString(KEY_USER_TOKEN, "")
+
+        if (apiKey == null || userId == null || userToken == null) {
```
JcMinarro, line 43: Because you are using a default parameter when ask for sharedPreferences keys, this condition will be false always. We can do something like:
```suggestion
        if (apiKey.isNullOrBlank() || userId.isNullOrBlank() || userToken.isNullOrBlank()) {
```

livedata/src/main/java/io/getstream/chat/android/livedata/service/sync/SyncService.kt
```diff
@@ -7,43 +7,48 @@ import android.content.Context
...
 import io.getstream.chat.android.livedata.ChatDomain
 
 class SyncService : Service() {
 
     override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
         createNotificationChannel()
+        showForegroundNotification()
 
+        intent?.getStringExtra(EXTRA_CID)?.let {
+            val config = EncryptedBackgroundSyncConfigStore(applicationContext).get()
```
JcMinarro, line 24: What happen if the returned config is `BackgroundSyncConfig.UNAVAILABLE`?
Shouldn't we stop the sinchronization process?

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
samiuelson: Both token and apiKey values are available in Chat.setUser(). 

The question is about ChatDomain.Builder API.

Now, ChaDomain is instantiated inside Chat.setUser() function (this is in UI SDK level):
```
ChatDomain.Builder(context, client, user)...build()
``` 
we can't get token and apiKey from client and User internally in ChatDomain (level) since those data are not visible. Those values can only be passed explicitly to the builder inside `Chat.setUser()` (in UI SDK).

So, on UI SDK level we have the new property in Builder: 
Chat.Builder. backgroundSyncEnabled: Boolean
This is fine.

But in livedata layer we need to pass apiKey and token:
```
ChatDomain.Builder.backgroundSyncEnabled(apiKey, user.getId(), userToken)
```
This function is called internally in UI SDK. 

ChatDomain.Builder.backgroundSyncEnabled() is not supposed to be internal function but a public function in ChatDomain.Builder allowing to specify if we want to have background sync enabled. If yes, it stores the values passed to allow sync. At the moment, there is no way to get those data internally from ChatDomain or User internally on livedata level.

samiuelson: To avoid passing those params in `ChatDomain.Builder.backgroundSyncEnabled(apiKey, user.getId()`
and to replace it with, for example: `ChatDomain.Builder.backgroundSyncEnabled: Boolean` we would need to pass apiKey and userToken somewhere else to the ChatDomain. 

They could be passed in constructor of ChatDomain.Builder(). Or they could be exposed by ChatClient and obtained from it (now they are not exposed).

livedata/src/main/java/io/getstream/chat/android/livedata/service/sync/BackgroundSyncConfigStore.kt
samiuelson: Maybe later if it's injected into some usecase to allow mocking. Now I'll remove it.

livedata/src/main/java/io/getstream/chat/android/livedata/service/sync/BackgroundSyncConfigStore.kt
samiuelson: This is just to satisfy lint which interprets result of getString() as String? type. 
Anyway, isNullOrBlank() is better here, so I'll use it

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
samiuelson: As discussed, I've moved the config storing logic outside the ChatDomainImpl's constructor. 
As for `ChatDomain.Builder.backgroundSyncEnabled()`, at the moment I end up with apiKey, userToken arguments.
I'm planning to refactor ChatClient later to expose those.

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
samiuelson: I've extracted EncryptedBackgroundSyncConfigStore creation into SyncProvider class. 
If you have a better concept of implementing DI in this project, we can discuss and refactor later.

PR #25, opened 2020-08-10
[CHORE] Update libraries

filipedelimabrito:
This PR updates some libraries used in this project.

It also changes from single quotes to double ones. Although both are acceptable, it is a good idea to use double quotes, especially when you are already using string interpolation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,27 +3,24 @@
 buildscript {
     ext {
         kotlin_version = '1.3.72'
-        jacocoVersion = '0.1.4'
-        junitVersion = '4.12'
-        jupiter_version = '5.5.1'
-        mannodermausVersion = '1.5.1.0'
+        mannodermaus_version = '1.5.1.0'
     }
     repositories {
         maven { url "https://plugins.gradle.org/m2/" }
         google()
         jcenter()
-
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.0'
+        classpath 'com.android.tools.build:gradle:3.6.4'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath 'com.dicedmelon.gradle:jacoco-android:0.1.4'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:2.1'
-        classpath "de.mannodermaus.gradle.plugins:android-junit5:$mannodermausVersion"
-        // NOTE: Do not place your application dependencies here; they belong
-        // in the individual module build.gradle files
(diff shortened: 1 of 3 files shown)


livedata/build.gradle
```diff
@@ -80,65 +78,64 @@ android {
...
     // optional - Kotlin Extensions and Coroutines support for Room
     implementation "androidx.room:room-ktx:$room_version"
-    implementation 'com.google.code.gson:gson:2.8.6'
+    implementation "com.google.code.gson:gson:2.8.6"
     implementation "androidx.room:room-ktx:$room_version"
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.5'
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.5'
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.8"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.6"
     // Kotlin + coroutines
```
filipedelimabrito, line 95: Is this comment useful?

JcMinarro: No, it isn't

livedata/build.gradle
```diff
@@ -23,17 +23,15 @@ jacocoAndroidUnitTestReport {
...
 }
 
-
 android {
-    compileSdkVersion 28
+    compileSdkVersion 29
     buildToolsVersion "29.0.3"
     defaultConfig {
         minSdkVersion 21
-        targetSdkVersion 28
```
JcMinarro, line 36: The SDK is not ready to be targeted to Android SDK Version 29 yet, we need to do some changes previously

build.gradle
```diff
@@ -3,27 +3,24 @@
 buildscript {
     ext {
         kotlin_version = '1.3.72'
-        jacocoVersion = '0.1.4'
-        junitVersion = '4.12'
-        jupiter_version = '5.5.1'
-        mannodermausVersion = '1.5.1.0'
+        mannodermaus_version = '1.5.1.0'
```
filipedelimabrito, line 6: Removed unused extra properties. Also changes from `mannodermausVersion` to `mannodermaus_version ` - always a good idea to have the same pattern (I could see something_version is being most used here, so the way to go now).

livedata/build.gradle
filipedelimabrito: @JcMinarro, it is now addressed by putting the version 28 again.
Thank you for reviewing.

30c5fa4

build.gradle
JcMinarro: Yes, it is fine for now. We are migrating it to a Kotlin file into the `buildSrc` project.

PR #26, opened 2020-08-10
[FIX] Typo

filipedelimabrito:
Now declared as `positiveRandomInt(..)` instead of `positveRandomInt(..)`.


livedata/src/test/java/io/getstream/chat/android/livedata/Mother.kt
```diff
@@ -9,28 +9,32 @@ import java.util.*
 import java.util.concurrent.ThreadLocalRandom
 
 private val charPool: List<Char> = ('a'..'z') + ('A'..'Z') + ('0'..'9')
-val random
-    get() = ThreadLocalRandom.current()
```
filipedelimabrito, line 13: There is no need to declare this custom accessor here. This declaration also can lead to nullability issues - which should be always avoided.

PR #19, opened 2020-07-22
Handle nullability properly on EventHandlerImpl

JcMinarro:
Acording with the [Event Object documentation](https://getstream.io/chat/docs/event_object/?language=js#event-object) Event's attributes are optionals. It force us to have null values on our event models and they aren't handled on the proper way.
We need to define a proper strcuture for our Events and then create our models with non-null values.
Until we have it defined, I have Null-Safe our code on `EventHandlerImpl`

Fix #17

PR #27, opened 2020-08-10
[FIX] NPE

### LINKED ISSUE #18: KotlinNullPointerException
SDK 4.2.11-beta-5

When the users join/leave channels, sometimes I retrieve the next crash:

image

It is not reproduced always, but rather often.

I saw the source code, and I think that using the "!!" operator is a bad idea, that can lead the project to crashes because it does not guarantee that variables are not null. I think you should do refactor these places. Thank you.

gz#4521

filipedelimabrito:
Closes #18 

Note: This prevents NPEs coming from some `event` properties and also logs them. I've checked the discussion [here](https://github.com/GetStream/stream-chat-android-livedata/pull/19) and to overcome skipping the updates you can probably implement a fallback strategy.


JcMinarro:
Thanks for PR.
With those changes we fix the NPE but hide some bad behavior on the SDK, and we want to avoid that.
That is why we are working into an updated on our [Client SDK](https://github.com/GetStream/stream-chat-android-client) to define a properly structure on our event classes

filipedelimabrito:
Thanks for letting me know @JcMinarro! 

Based on that, I am going to close this PR. Once the structure is defined I'll be glad to help again.

PR #29, opened 2020-08-19
Feautre/cas 175 notification displaying

samiuelson:
* Serialize and store notification config
* Create dedicated firebase service with offline sync support


livedata/build.gradle
```diff
@@ -78,7 +78,7 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
 
-    api "com.github.GetStream:stream-chat-android-client:1.10.0"
+    api 'com.github.GetStream:stream-chat-android-client:1.14.0-SNAPSHOT'
```
JcMinarro, line 81: Version 1.14.0 is already released ;)

livedata/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessageService.kt
```diff
@@ -0,0 +1,147 @@
...
+import kotlinx.coroutines.launch
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+
+class OfflineSyncFirebaseMessagingService : FirebaseMessagingService() {
+
+    private val logger = ChatLogger.get("OfflineSyncFirebaseMessagingService")
+
+    private val syncModule by lazy {
```
JcMinarro, line 28: Do we need this class?
It seems only a holder for `EncryptedBackgroundSyncConfigStore` and `NotificationConfigStore` that don't need anything else to be built, they could be here instead

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -148,6 +157,15 @@ interface ChatDomain {
                 }
             }
         }
+
+        private fun storeNotificationConfig(notificationConfig: NotificationConfig) {
+            if (NotificationConfigUnavailable != notificationConfig) {
+                syncModule.notificationConfigStore.apply {
+                    clear()
```
JcMinarro, line 164: Is it needed to "clear" before "put" a new value?
`put()` methods normally override previous value

JcMinarro (approved):

PR #30, opened 2020-08-20
Update ChatEvent structure

### LINKED ISSUE #18: KotlinNullPointerException

JcMinarro:
`ChatEvent` has been refactored on LLC.
Now we have a sealed class with different events. Every Event has only the required fields, so no more nullable values.
This PR will fix #18


tschellenbach:
Great improvement!

Issue #31, opened 2020-08-25
message order when offline


tschellenbach:
retryMessage/ selectSyncNeeded don't implement sorting. So if you:

- go offline
- send 10 messages
- go online

it won't show up in the right order.

PR #34, opened 2020-08-26
Bugfix/message retry sort

tschellenbach:
1. update the test to fail if we return message in the wrong order
2. add the "ORDER BY" in the query

Fixes: https://github.com/GetStream/stream-chat-android-livedata/issues/31


JcMinarro (approved):
Look Good!!!

PR #35, opened 2020-08-27
[cas-217] Fix command support

samiuelson:
* Make the `runAndRetry()` function type-safe
* Update the message in storage with the one processed by the backend 
* Emmit a backend-processed message event to channel once "send message" operation was successful 
Without ☝🏼  the giphy picker couldn't be rendered because the emitted message didn't contain `type="ephemeral"` and `command="giphy"` values which are being set by the backend.

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -270,9 +270,9 @@ class ChatDomainImpl private constructor(
         return database
     }
 
-    suspend fun runAndRetry(runnable: () -> Call<*>): Result<*> {
+    suspend fun <T> runAndRetry(runnable: () -> Call<T>): Result<T> {
         var attempt = 1
-        var result: Result<*>
+        var result: Result<T>
 
         while (true) {
             result = runnable().execute()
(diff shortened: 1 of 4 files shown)


tschellenbach:
So the previous behaviour was:

- insert local message
- mark local message completed
- event from message.new updates with the backend message

new behaviour:

- insert local message
- take the message response and mark it completed
- event from message.new updates with the backend message

I don't understand yet how the old behaviour prevents the giphy command from working?

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -270,9 +270,9 @@ class ChatDomainImpl private constructor(
         return database
     }
 
-    suspend fun runAndRetry(runnable: () -> Call<*>): Result<*> {
+    suspend fun <T> runAndRetry(runnable: () -> Call<T>): Result<T> {
```
tschellenbach, line 273: 👍

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -405,35 +404,34 @@ class ChannelControllerImpl(
...
-            }
-            val result = domainImpl.runAndRetry(runnable)
+            val result = domainImpl.runAndRetry { channelController.sendMessage(message) }
             if (result.isSuccess) {
-                // set sendMessageCompletedAt so we know when to edit vs call sendMessage
-                messageEntity.syncStatus = SyncStatus.COMPLETED
-                messageEntity.sendMessageCompletedAt = Date()
-                domainImpl.repos.messages.insert(messageEntity)
-                output = Result(result.data() as Message?, null)
+                val processedMessage: Message = result.data()
```
tschellenbach, line 412: 👍

JcMinarro (approved):

Issue #32, opened 2020-08-26
offline reactions don't trigger livedata update


tschellenbach:
one issue is that we don't use immutable type, which can break diffutils/livedata

but I don't think this is the main issue here, there is a second and perhaps more important issue:

- ReactionDialogAdapter.java hasn't been converted to Kotlin yet
https://github.com/GetStream/stream-chat-android/blob/master/library/src/main/java/com/getstream/sdk/chat/adapter/ReactionDialogAdapter.java

- cid is specified as a string, which can be ""
- when it's "" reactions don't work
- the cid is empty since message.channel and message.cid are not defined
- the channel object is very large, in this case we only want to pass a channel id, channel reference type of object. 

There are a few possible solutions:

- pass the channel separately from the message to the adapter. (perhaps remove this field from the message if it's not specified.) 
- automatically set message.channel on the message object from livedata (don't really like this since the channel object has so much info.)

Discussion:

- Do we need a message.channel field? (I don't see where it's used)
- Do we need a message.cid field? (again, I don't think it's used.)

@JcMinarro and @tbarbugli thoughts on this?

tschellenbach:
the UI in stream-chat-android is using reactionCounts

![image](https://user-images.githubusercontent.com/265409/91496561-91060880-e879-11ea-9854-698894a7a9ef.png)

tschellenbach:
confirmed, if i add the cid on the message reactions work...

tschellenbach:
first issue also exists, see the message object (different) and reaction count (same)

![image](https://user-images.githubusercontent.com/265409/92829631-a870e580-f391-11ea-9949-f0cccd3674b3.png)

tschellenbach:
fixed here: https://github.com/GetStream/stream-chat-android-livedata/pull/36

PR #36, opened 2020-08-27
Bugfix/reactions

tschellenbach:
There are a few bugs with reactions that this PR solves.

**Mutability**

Adding reactions is somewhat bugged right now since we modify objects in place. The best practice is to have immutability. (similar to how React/RN etc handle this).

- So we should use val instead of var & list instead of mutableList etc.

To fix this, this PR

- copies the mutable data structures on the message and modifies the copy
- creates a copy of the message object

This is a patch, the better solution is to update our data structure to be immutable. 

Related to https://github.com/GetStream/stream-chat-android-livedata/issues/32

**Composite builds**

Similar to this PR, add support for composite builds: https://github.com/GetStream/stream-chat-android-client/pull/71

**Removing reactions**

This code had 2 bugs. 1. It didn't correctly remove objects in some cases. 2. The reaction counts were set to 0 instead of removed which messed with the UI

**Diff Callback**

The diff callback didn't compare objects correctly. Moved it to Kotlin and fixed that.


livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -64,20 +46,29 @@ fun Channel.users(): List<User> {
...
     }
 
     // add to latest reactions
+    this.latestReactions = this.latestReactions.toMutableList()
     this.latestReactions.add(reaction)
 
     // update the count
     val currentCount = this.reactionCounts.getOrElse(reaction.type) { 0 }
+    // copy the object so livedata's diffutils can notice a change
+    this.reactionCounts = this.reactionCounts.toMutableMap()
```
JcMinarro, line 64: You are mutating `this` (on that case `message`) so the `equals()` method over `message` won't work either

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -691,7 +692,7 @@ class ChannelControllerImpl(
 
         // update all the fresh messages
         for (message in freshMessages) {
-            copy[message.id] = message
+            copy[message.id] = message.copy()
```
JcMinarro, line 695: Because we have a mutable model and we are not updating anything, I think it is not needed to call `.copy()` method here

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -97,14 +97,15 @@ class ChannelControllerImpl(
...
+    override val messages: LiveData<List<Message>> = Transformations.map(_messages) { messageMap ->
         // TODO: consider removing this check
-        it.values
+        val messages = messageMap.values
             .asSequence()
             .filter { it.parentId == null || it.showInChannel }
             .filter { hideMessagesBefore == null || it.createdAt!! > hideMessagesBefore }
             .sortedBy { it.createdAt }
             .toList()
+        messages.map { it.copy(cid = cid) }
```
JcMinarro, line 108:
```suggestion
        messageMap.values
            .asSequence()
            .filter { it.parentId == null || it.showInChannel }
            .filter { hideMessagesBefore == null || it.createdAt!! > hideMessagesBefore }
            .sortedBy { it.createdAt }
            .toList()
            .map { it.copy(cid = cid) }
```
We can join both lines

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
```diff
@@ -83,28 +83,39 @@ data class MessageEntity(@PrimaryKey var id: String, var cid: String, var userId
         }
 
         // add to latest reactions
+        latestReactions = latestReactions.toMutableList()
```
JcMinarro, line 86: Because `latestReaction` is already a MutableList, this line shouldn't be needed

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
```diff
@@ -83,28 +83,39 @@ data class MessageEntity(@PrimaryKey var id: String, var cid: String, var userId
...
         latestReactions.add(reactionEntity)
 
         // update the count
+        reactionCounts = reactionCounts.toMutableMap()
```
JcMinarro, line 90: Same here

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
```diff
@@ -83,28 +83,39 @@ data class MessageEntity(@PrimaryKey var id: String, var cid: String, var userId
...
         val currentCount = reactionCounts.getOrElse(reaction.type) { 0 }
         reactionCounts[reaction.type] = currentCount + 1
         // update the score
+        reactionScores = reactionScores.toMutableMap()
```
JcMinarro, line 94: And here

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
```diff
@@ -83,28 +83,39 @@ data class MessageEntity(@PrimaryKey var id: String, var cid: String, var userId
...
+        ownReactions = ownReactions.filterNot { it.type == reactionEntity.type && it.userId == reactionEntity.userId }.toMutableList()
+        latestReactions = latestReactions.filterNot { it.type == reactionEntity.type && it.userId == reactionEntity.userId }.toMutableList()
+        val countAfterFilter = ownReactions.size + latestReactions.size
         if (updateCounts) {
-            val shouldDecrement = removed1 || removed2 || latestReactions.size >= 15
+            val shouldDecrement = (countBeforeFilter > countAfterFilter) || latestReactions.size >= 15
             if (shouldDecrement) {
                 val currentCount = reactionCounts.getOrElse(reaction.type) { 1 }
                 reactionCounts[reaction.type] = currentCount - 1
+                if (reactionCounts[reaction.type] == 0) {
```
JcMinarro, line 111:
```suggestion
                if (reactionCounts[reaction.type] <= 0) {
```
Shouldn't be `<= 0`?

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
```diff
@@ -83,28 +83,39 @@ data class MessageEntity(@PrimaryKey var id: String, var cid: String, var userId
...
+                    reactionCounts.remove(reaction.type)
+                }
                 val currentScore = reactionScores.getOrElse(reaction.type) { 1 }
                 reactionScores[reaction.type] = currentScore - reaction.score
+                if (reactionScores[reaction.type] == 0) {
```
JcMinarro, line 116:
```suggestion
                if (reactionScores[reaction.type] <= 0) {
```
Shouldn't be `<= 0`?

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -46,33 +46,50 @@ fun Channel.users(): List<User> {
 }
 
 fun Message.addReaction(reaction: Reaction, isMine: Boolean) {
+
     // add to own reactions
     if (isMine) {
+        this.ownReactions = this.ownReactions.toMutableList()
```
JcMinarro, line 52: Same here

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -46,33 +46,50 @@ fun Channel.users(): List<User> {
...
         this.ownReactions.add(reaction)
     }
 
     // add to latest reactions
+    this.latestReactions = this.latestReactions.toMutableList()
```
JcMinarro, line 57: Same here

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -46,33 +46,50 @@ fun Channel.users(): List<User> {
...
+    val countAfterFilter = ownReactions.size + latestReactions.size
+
     if (updateCounts) {
-        val shouldDecrement = removed1 || removed2 || this.latestReactions.size >= 15
+        val shouldDecrement = countBeforeFilter > countAfterFilter || this.latestReactions.size >= 15
         if (shouldDecrement) {
+            this.reactionCounts = this.reactionCounts.toMutableMap()
             val currentCount = this.reactionCounts.getOrElse(reaction.type) { 1 }
             this.reactionCounts[reaction.type] = currentCount - 1
+            if (reactionCounts[reaction.type] == 0) {
```
JcMinarro, line 84: Same here

tschellenbach:
toMutableMap and .copy() create a copy of the object. This prevents issues where modifying the same object causes diffutils not to pick up changes since both old & new are modified. Copy on the message creates a shallow copy (ie not a deepcopy) so we also need to copy the reactionCounts and similar fields. Longer term solution is to make the Message and Channel classes immutable.

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
tschellenbach: yes, that's a more reliable approach

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
tschellenbach: calling toMutableList creates a copy of the object. this ensures that we're not modifying the same object in place.

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
tschellenbach: 👍

PR #38, opened 2020-09-04
[cas-226] giphy shuffle, cancel support

samiuelson:
Add usecases to support giphy picker shuffle, cancel support

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
@@ -30,6 +30,7 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import java.io.File
+import java.lang.IllegalArgumentException
 import java.util.Calendar
 import java.util.Date
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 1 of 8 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/usecase/UseCaseHelper.kt
```diff
@@ -66,6 +66,24 @@ data class UseCaseHelper(var chatDomainImpl: ChatDomainImpl) {
      * The API call to create the message is retried using the retry policy
      */
     var sendMessage: SendMessage = SendMessageImpl(chatDomainImpl)
+    /**
+     * Cancel an emphemeral message. This message is immediately removed from local storage.
+     * The API call to delete the message is retried using the retry policy
+     */
+    var cancelMessage: CancelMessage = CancelMessage(chatDomainImpl)
```
JcMinarro, line 73: Those properties can be inmutable

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -575,7 +672,7 @@ class ChannelControllerImpl(
         return message
     }
 
-    fun upsertMessages(messages: List<Message>) {
+    private fun upsertMessages(messages: List<Message> = emptyList()) {
```
JcMinarro, line 675: Which case do we need to call to this method without argument and then use an emptyList?
If I am not wrong, this method won't do anything if the `messagesList` is empty. Am I wrong?

livedata/src/main/java/io/getstream/chat/android/livedata/usecase/UseCaseHelper.kt
samiuelson: making all the props of this class immutable

JcMinarro (approved):

PR #40, opened 2020-09-12
Fix/cas 188 remove duplicates of new channels

ogkuzmin:
When create a new channel it doubled in the channel list view. It happened because we didn't check duplicates of channels in QueryChannelsControllerImpl. The fix is very easy.
I added test for addChannel case.
When I was writing test I found that this class is coupled with concrete classes not interfaces, code is too coupled. Good point for refactoring

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
@@ -199,9 +199,9 @@ class QueryChannelsControllerImpl(
     private fun addChannels(newChannels: List<Channel>, onTop: Boolean = false) {
         // second page adds to the list of channels
         queryEntity.channelCids = if (onTop) {
-            newChannels.map { it.cid } + queryEntity.channelCids
+            (newChannels.map { it.cid } + queryEntity.channelCids).distinct()
         } else {
-            queryEntity.channelCids + newChannels.map { it.cid }
+            (queryEntity.channelCids + newChannels.map { it.cid }).distinct()
         }
 
         _channels.postValue(
--- a/livedata/src/test/java/androidx/arch/core/executor/testing/InstantExecutorExtension.kt
+++ b/livedata/src/test/java/androidx/arch/core/executor/testing/InstantExecutorExtension.kt
@@ -1,0 +1,25 @@
+package androidx.arch.core.executor.testing
+
+import androidx.arch.core.executor.ArchTaskExecutor
+import androidx.arch.core.executor.TaskExecutor
+import org.junit.jupiter.api.extension.AfterEachCallback
+import org.junit.jupiter.api.extension.BeforeEachCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+
+class InstantExecutorExtension : BeforeEachCallback, AfterEachCallback {
+
+    override fun beforeEach(context: ExtensionContext?) {
+        ArchTaskExecutor.getInstance()
+            .setDelegate(object : TaskExecutor() {
+                override fun executeOnDiskIO(runnable: Runnable) = runnable.run()
+
+                override fun postToMainThread(runnable: Runnable) = runnable.run()
+
+                override fun isMainThread(): Boolean = true
+            })
+    }
+
+    override fun afterEach(context: ExtensionContext?) {
+        ArchTaskExecutor.getInstance().setDelegate(null)
+    }
+}
--- a/livedata/src/test/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImplTest.kt
+++ b/livedata/src/test/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImplTest.kt
@@ -5,6 +5,7 @@ import com.nhaarman.mockitokotlin2.*
 import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.livedata.ChatDomainImpl
+import io.getstream.chat.android.livedata.randomChannel
 import io.getstream.chat.android.livedata.utils.getOrAwaitValue
 import kotlinx.coroutines.Job
 import org.amshove.kluent.shouldBeEqualTo
@@ -20,7 +21,7 @@ internal class QueryChannelsControllerImplTest {
             .givenNewChannelController(channelController)
             .setupChatControllersInstantiation()
             .get()
-        val newChannel = Channel(cid = "CID", id = "ID")
+        val newChannel = randomChannel()
 
         sut.addChannelIfFilterMatches(newChannel)
 
@@ -33,7 +34,7 @@ internal class QueryChannelsControllerImplTest {
             .givenNewChannelController(mock())
             .setupChatControllersInstantiation()
             .get()
-        val newChannel = Channel(cid = "ChannelType:ChannelID", id = "ChannelID")
+        val newChannel = randomChannel(cid = "ChannelType:ChannelID")
 
         sut.addChannelIfFilterMatches(newChannel)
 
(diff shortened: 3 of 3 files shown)


livedata/src/test/java/androidx/arch/core/executor/testing/InstantExecutorExtension.kt
```diff
@@ -0,0 +1,25 @@
+package androidx.arch.core.executor.testing
+
+import androidx.arch.core.executor.ArchTaskExecutor
+import androidx.arch.core.executor.TaskExecutor
+import org.junit.jupiter.api.extension.AfterEachCallback
+import org.junit.jupiter.api.extension.BeforeEachCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+
+class InstantExecutorExtension : BeforeEachCallback, AfterEachCallback {
```
zsmb13, line 9: Nicely done 👍

livedata/src/test/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImplTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        sut.addChannelIfFilterMatches(newChannel)
+        sut.addChannelIfFilterMatches(newChannel)
+
+        val result = sut.channels.getOrAwaitValue()
+        result.size shouldBeEqualTo 1
+        result.first().cid shouldBeEqualTo "ChannelType:ChannelID"
+    }
+}
+
+class Fixture {
```
zsmb13, line 62: I think this could be `private` within this file, if it's only meant for use in this one test

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -199,9 +199,9 @@ class QueryChannelsControllerImpl(
     private fun addChannels(newChannels: List<Channel>, onTop: Boolean = false) {
         // second page adds to the list of channels
         queryEntity.channelCids = if (onTop) {
-            newChannels.map { it.cid } + queryEntity.channelCids
+            (newChannels.map { it.cid } + queryEntity.channelCids).distinct()
         } else {
-            queryEntity.channelCids + newChannels.map { it.cid }
+            (queryEntity.channelCids + newChannels.map { it.cid }).distinct()
```
zsmb13, line 204: In the first branch, any newly added channels that were already present in the list will move to the top (because `distinct` will remove their second occurrence, which is their original position). That's intentional, right?

In the second branch, we're simply adding fewer channels to the end of the list this way, and the existing ones don't get modified / moved around.

zsmb13 (changes requested):
Looks good in general, had just a couple small questions.  

The test setup / structure is especially great, it was easy to understand what's happening there 👍

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
ogkuzmin: I can't answer this question properly from my current level of knowledge how the system should works. 
Doubles could happen when you had created new channel locally and then received event from backend that channel was created (really the same one). Usually it is already on top of the channel list. 
And if the order of such duplication really matters there is must be test for that.

livedata/src/test/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImplTest.kt
ogkuzmin: You're right

livedata/src/test/java/androidx/arch/core/executor/testing/InstantExecutorExtension.kt
ogkuzmin: I took it from view models library. One point more to keep sources in one repo

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
ogkuzmin: @JcMinarro what do you think?

JcMinarro: I think it is fine, When some new channel is added to the top of the list it is because new event is received (new channel created, channel receive a new message and so on). We are getting the id to perform a query over it (Some sync on background) and we want to keeps this order

livedata/src/test/java/androidx/arch/core/executor/testing/InstantExecutorExtension.kt
JcMinarro: [Testfixture](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures) is a great feature that Gradle provides us to handle it, but it is not supported by AGP yet :cry:  ([Jetbrains issue](https://youtrack.jetbrains.com/issue/KT-33877), [AGP Issue](https://issuetracker.google.com/issues/139438142), [AS Issue](https://issuetracker.google.com/issues/139762443))

livedata/src/test/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImplTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        sut.addChannelIfFilterMatches(newChannel)
+        sut.addChannelIfFilterMatches(newChannel)
+
+        val result = sut.channels.getOrAwaitValue()
+        result.size shouldBeEqualTo 1
+        result.first().cid shouldBeEqualTo "ChannelType:ChannelID"
+    }
+}
+
+private class Fixture {
```
JcMinarro, line 62: :+1: 
Good way to init your tests preconditions!!!

livedata/src/test/java/androidx/arch/core/executor/testing/InstantExecutorExtension.kt
zsmb13: ^ I think this comment is on the wrong discussion thread 🙂

zsmb13 (dismissed):

livedata/src/test/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImplTest.kt
```diff
@@ -0,0 +1,91 @@
...
+@ExtendWith(InstantExecutorExtension::class)
+internal class QueryChannelsControllerImplTest {
+    @Test
+    fun `when add channel if filter matches should update LiveData from channel to channel controller`() {
+        val channelController = mock<ChannelControllerImpl>()
+        val sut = Fixture()
+            .givenNewChannelController(channelController)
+            .setupChatControllersInstantiation()
+            .get()
+        val newChannel = Channel(cid = "CID", id = "ID")
```
JcMinarro, line 23: I created a [`Mother` class](https://github.com/GetStream/stream-chat-android-livedata/blob/46cb039c3258548f937d169443449f4854a5f182/livedata/src/test/java/io/getstream/chat/android/livedata/Mother.kt#L140) that helps us creating object. Whatever new parameter is added to this entity, will be handled there, so your tests won't break if a new mandatory field is added.
On the other hand, because it use random values, we could detect if any field cause our tests fail (maybe internally we are using some flag/field that has not been considered on our tests).
Those "Mother methods" use the powerfull of Kotlin to provide default values but allowing us to set our needed attributes when we need some custom object

ogkuzmin: You're right. Mother is very convenient and useful pattern in unit tests

JcMinarro (approved):

PR #43, opened 2020-09-15
Fix/cas 240 crash create channel offline

ogkuzmin:
1) Fix. Set author of channel if one was not set when create a new channel
2) Added test for this case
3) Refactor ChatDomainImpl in order to have possibility to write unit tests


tschellenbach:
Why do you need a factory pattern to facilitate unit testing?

ogkuzmin:
Only for the reason to create Handler, DB instances and provide them directly to constructor of Domain

PR #45, opened 2020-09-16
Fix/cas 243 fix channel db mapping

ogkuzmin:
Tried to fix bug https://console.firebase.google.com/project/stream-chat-internal/crashlytics/app/android:io.getstream.chat.sample.debug/issues/e6bdc1263924f1ec3ac8a897d636c231?time=last-ninety-days&sessionId=5F5BB9D001150001483DD273418A3C65_DNE_0_v2
Couldn't reproduce. There are two possible points:
### 1) ChannelRepository::retryChannels. 
For me there was not any point to map to Channel class in the repo layer. It could lead to such exception.
### 2) ChatDomainImpl::selectAndEnrichChannels.
The logic seems correct. Just rewrite this place to get all users for channels in order to bring power of Kotlin functionality to work with collections


JcMinarro (approved):
It looks fine, 
Seems that there aren't any reason con convert from `ChannelEntity` to `Channel` at that point.
BTW, we have some tasks to refactor our "entity mappers", maybe it is a good task that you can handle in the future ;)

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -616,31 +616,22 @@ class ChatDomainImpl private constructor(
...
-                channelMessagesMap[channelEntity.cid] = messages
-            }
+        val userIds = channelEntities.fold(emptySet<String>()) { acc, channel ->
+            acc + channel.createdByUserId.orEmpty() +
+                channel.members.keys +
+                channel.reads.keys +
+                channelMessagesMap[channel.cid]?.map { message ->
+                    message.userId + message.latestReactions.map(ReactionEntity::userId)
+                }.orEmpty()
         }
```
tschellenbach, line 636: what a nifty kotlin feature

livedata/src/main/java/io/getstream/chat/android/livedata/repository/ChannelRepository.kt
```diff
@@ -66,13 +71,12 @@ class ChannelRepository(var channelDao: ChannelDao, var cacheSize: Int = 100, va
...
     suspend fun retryChannels(): List<ChannelEntity> {
-        val userMap: Map<String, User> = mutableMapOf(currentUser.id to currentUser)
         val channelEntities = selectSyncNeeded()
 
         for (channelEntity in channelEntities) {
-            val channel = channelEntity.toChannel(userMap)
             // TODO: what about channel.members
-            val result = client.createChannel(channel.type, channel.id, channel.extraData).execute()
+            val result =
+                client.createChannel(channelEntity.type, channelEntity.channelId, channelEntity.extraData).execute()
```
tschellenbach, line 79: The stack trace is for a different code path

image

tschellenbach:
Somehow a channel is stored in offline state with a value for the read state, that doesn't include the user.

The read state comes from either:
- an event
- query Channel API call
- query Channels API call

I suspect that one of these 3 is not storing the user information which causes this error.

tschellenbach:
1. See EventHandlerImpl and have a look at the MessageReadEvent and NotificationMarkReadEvent it seems to me that neither correctly stores the user object

image

missing something like `users[event.user.id] = UserEntity(event.user)`

image



2. See ChatDomainImpl storeStateForChannels, this seems fine

note how it calls channel.users()

image

and channel.users stores users for read

image

PR #47, opened 2020-09-17
Cas 243/fix storing users in db

ogkuzmin:
According to Thierry's notes in #45 there are other possible places where we could drop required users
It seems we forgot to keep users for MessageReadEvent and NotificationMarkReadEvent.
1) Fix is very easy, collect user from such event
2) It's easy to make such a mistake and forget to do something or keep something. I saw in the events hierarchy there is CidEvent class that has only the cid field. In these changes there is an example how to keep CID from all this event without code duplication. If you find it's good we can create such things for other field (example with field user)

``` kotlin
interface EventWithUser /* or other name */ {
  val user: User
}

...

// and keep all users from events

events.filterIsInstance<EventWithUser>().onEach { users += it.user.id to it.user }
```

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
@@ -1,7 +1,59 @@
 package io.getstream.chat.android.livedata
 
 import exhaustive
-import io.getstream.chat.android.client.events.*
+import io.getstream.chat.android.client.events.ChannelCreatedEvent
+import io.getstream.chat.android.client.events.ChannelDeletedEvent
+import io.getstream.chat.android.client.events.ChannelHiddenEvent
+import io.getstream.chat.android.client.events.ChannelMuteEvent
+import io.getstream.chat.android.client.events.ChannelTruncatedEvent
+import io.getstream.chat.android.client.events.ChannelUnmuteEvent
+import io.getstream.chat.android.client.events.ChannelUpdatedEvent
+import io.getstream.chat.android.client.events.ChannelUserBannedEvent
+import io.getstream.chat.android.client.events.ChannelUserUnbannedEvent
+import io.getstream.chat.android.client.events.ChannelVisibleEvent
+import io.getstream.chat.android.client.events.ChannelsMuteEvent
+import io.getstream.chat.android.client.events.ChannelsUnmuteEvent
+import io.getstream.chat.android.client.events.ChatEvent
+import io.getstream.chat.android.client.events.CidEvent
+import io.getstream.chat.android.client.events.ConnectedEvent
+import io.getstream.chat.android.client.events.ConnectingEvent
+import io.getstream.chat.android.client.events.DisconnectedEvent
+import io.getstream.chat.android.client.events.ErrorEvent
+import io.getstream.chat.android.client.events.GlobalUserBannedEvent
+import io.getstream.chat.android.client.events.GlobalUserUnbannedEvent
+import io.getstream.chat.android.client.events.HealthEvent
+import io.getstream.chat.android.client.events.MemberAddedEvent
+import io.getstream.chat.android.client.events.MemberRemovedEvent
+import io.getstream.chat.android.client.events.MemberUpdatedEvent
+import io.getstream.chat.android.client.events.MessageDeletedEvent
+import io.getstream.chat.android.client.events.MessageReadEvent
+import io.getstream.chat.android.client.events.MessageUpdatedEvent
+import io.getstream.chat.android.client.events.NewMessageEvent
+import io.getstream.chat.android.client.events.NotificationAddedToChannelEvent
+import io.getstream.chat.android.client.events.NotificationChannelDeletedEvent
+import io.getstream.chat.android.client.events.NotificationChannelMutesUpdatedEvent
+import io.getstream.chat.android.client.events.NotificationChannelTruncatedEvent
+import io.getstream.chat.android.client.events.NotificationInviteAcceptedEvent
+import io.getstream.chat.android.client.events.NotificationInvitedEvent
+import io.getstream.chat.android.client.events.NotificationMarkReadEvent
+import io.getstream.chat.android.client.events.NotificationMessageNewEvent
+import io.getstream.chat.android.client.events.NotificationMutesUpdatedEvent
+import io.getstream.chat.android.client.events.NotificationRemovedFromChannelEvent
+import io.getstream.chat.android.client.events.ReactionDeletedEvent
+import io.getstream.chat.android.client.events.ReactionNewEvent
+import io.getstream.chat.android.client.events.ReactionUpdateEvent
+import io.getstream.chat.android.client.events.TypingStartEvent
+import io.getstream.chat.android.client.events.TypingStopEvent
+import io.getstream.chat.android.client.events.UnknownEvent
+import io.getstream.chat.android.client.events.UserDeletedEvent
+import io.getstream.chat.android.client.events.UserMutedEvent
+import io.getstream.chat.android.client.events.UserPresenceChangedEvent
+import io.getstream.chat.android.client.events.UserStartWatchingEvent
+import io.getstream.chat.android.client.events.UserStopWatchingEvent
+import io.getstream.chat.android.client.events.UserUnmutedEvent
+import io.getstream.chat.android.client.events.UserUpdatedEvent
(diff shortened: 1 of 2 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -24,26 +21,9 @@ fun Message.getCid(): String =
...
-    users.add(this.createdBy)
-    for (member in this.members) {
-        users.add(member.user)
-    }
-    for (read in this.read) {
-        users.add(read.user)
-    }
-    return users
-}
+fun Channel.users() = members.map(Member::user) + read.map(ChannelUserRead::user) + createdBy
```
tschellenbach, line 26: had no idea Kotlin could do this, cool

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -221,6 +142,7 @@ class EventHandlerImpl(var domainImpl: ChatDomainImpl, var runAsync: Boolean = t
                 }
                 is MessageReadEvent -> {
                     // get the channel, update reads, write the channel
+                    users[event.user.id] = UserEntity(event.user)
```
tschellenbach, line 145: pretty sure this fixes the crash 👏

tschellenbach:
Looks good to me, but nice if @JcMinarro can also take a look. Especially the filter refactor I wasn't 100% sure about given the talk about how it modifies in place.

tschellenbach (approved):

tschellenbach:
Also good to verify that user watching and user stop watching don't store the user in the DB. otherwise a livestream type of use case could cause some issues with offline storage.

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -89,25 +37,24 @@ class EventHandlerImpl(var domainImpl: ChatDomainImpl, var runAsync: Boolean = t
...
+                is MemberUpdatedEvent, is ChannelUpdatedEvent, is ChannelDeletedEvent, is ChannelHiddenEvent,
+                is ChannelVisibleEvent, is NotificationAddedToChannelEvent, is NotificationInvitedEvent,
+                is NotificationInviteAcceptedEvent, is ChannelTruncatedEvent, is ChannelCreatedEvent, is HealthEvent,
+                is NotificationMutesUpdatedEvent, is GlobalUserBannedEvent, is UserDeletedEvent, is UserMutedEvent,
+                is UsersMutedEvent, is UserPresenceChangedEvent, is GlobalUserUnbannedEvent, is UserUnmutedEvent,
+                is UsersUnmutedEvent, is UserUpdatedEvent, is NotificationChannelMutesUpdatedEvent, is ConnectedEvent,
+                is ConnectingEvent, is DisconnectedEvent, is ErrorEvent, is UnknownEvent, is NotificationMessageNewEvent,
+                is NotificationChannelDeletedEvent, is NotificationChannelTruncatedEvent, is NotificationMarkReadEvent,
+                is TypingStartEvent, is TypingStopEvent, is ChannelUserBannedEvent, is UserStartWatchingEvent,
+                is UserStopWatchingEvent, is ChannelUserUnbannedEvent -> Unit
```
JcMinarro, line 55: It is good to group all "empty cases" into one statement, but, could you keep them in different lines?
It is easier to read and will help us minimizing conflict on futures changes

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -24,26 +21,9 @@ fun Message.getCid(): String =
...
 
-fun Message.users(): List<User> {
-    val users = mutableListOf<User>()
-    users.add(this.user)
-    for (reaction in this.latestReactions) {
-        reaction.user?.let { users.add(it) }
-    }
-    return users
-}
+fun Message.users() = latestReactions.mapNotNull(Reaction::user) + user
```
JcMinarro, line 24: Those extensions functions should be `internal`

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -1,59 +1,7 @@
...
-import io.getstream.chat.android.client.events.UserDeletedEvent
-import io.getstream.chat.android.client.events.UserMutedEvent
-import io.getstream.chat.android.client.events.UserPresenceChangedEvent
-import io.getstream.chat.android.client.events.UserStartWatchingEvent
-import io.getstream.chat.android.client.events.UserStopWatchingEvent
-import io.getstream.chat.android.client.events.UserUnmutedEvent
-import io.getstream.chat.android.client.events.UserUpdatedEvent
-import io.getstream.chat.android.client.events.UsersMutedEvent
-import io.getstream.chat.android.client.events.UsersUnmutedEvent
+import io.getstream.chat.android.client.events.*
```
JcMinarro, line 4: We shouldn't use wildcard imports in our code.
It is a rule on KtLint and I was a little surprised why it was passing our CI process.
At the end I have seen there is a [here](https://github.com/GetStream/stream-chat-android-livedata/blob/master/.editorconfig#L2) but it will be removed, so, could you revert this change?

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -365,13 +288,8 @@ class EventHandlerImpl(var domainImpl: ChatDomainImpl, var runAsync: Boolean = t
...
                 }
-                is TypingStartEvent -> { }
-                is TypingStopEvent -> { }
-                is HealthEvent -> { }
-                is ConnectingEvent -> { }
-                is DisconnectedEvent -> { }
-                is ErrorEvent -> { }
-                is UnknownEvent -> { }
+                is TypingStartEvent, is TypingStopEvent, is HealthEvent, is ConnectingEvent, is DisconnectedEvent,
+                is ErrorEvent, is UnknownEvent -> Unit
```
JcMinarro, line 292: Same here

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -3,10 +3,7 @@ package io.getstream.chat.android.livedata.extensions
 import io.getstream.chat.android.client.api.models.QuerySort
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.errors.ChatNetworkError
-import io.getstream.chat.android.client.models.Channel
-import io.getstream.chat.android.client.models.Message
-import io.getstream.chat.android.client.models.Reaction
-import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.client.models.*
```
JcMinarro, line 6: Same here

JcMinarro (changes requested):

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
ogkuzmin: fix

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
ogkuzmin: did it

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
ogkuzmin: Fixed

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
ogkuzmin: Fixed

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
ogkuzmin: Fixed

JcMinarro (approved):
LGTM

PR #48, opened 2020-09-18
Update event subscriptions for new API

zsmb13:
API usage updates, to be merged after https://github.com/GetStream/stream-chat-android-client/pull/76 is merged and released


livedata/build.gradle
```diff
@@ -68,7 +68,7 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
 
-    api "com.github.getstream:stream-chat-android-client:1.15.5"
+    api "com.github.getstream:stream-chat-android-client:1.15.6"
```
zsmb13, line 71: This doesn't exist yet, so the pipeline will be broken for now. Can be tested locally by publishing the new version of the LLC from the appropriate branch and depending on that.

JcMinarro (approved):

PR #49, opened 2020-09-18
Update ktlint rules

JcMinarro:
Update Ktlint

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/converter/ListConverter.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/converter/ListConverter.kt
@@ -5,7 +5,6 @@ import com.google.gson.reflect.TypeToken
 import io.getstream.chat.android.client.models.Attachment
 import io.getstream.chat.android.livedata.entity.ReactionEntity
 import io.getstream.chat.android.livedata.gson
-import java.util.*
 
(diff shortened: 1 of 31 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/converter/ListConverter.kt
```diff
@@ -5,7 +5,6 @@ import com.google.gson.reflect.TypeToken
 import io.getstream.chat.android.client.models.Attachment
 import io.getstream.chat.android.livedata.entity.ReactionEntity
 import io.getstream.chat.android.livedata.gson
-import java.util.*
```
zsmb13, line 8: Wonder how this wasn't removed by Studio at some point 🤔

zsmb13 (approved):

livedata/src/main/java/io/getstream/chat/android/livedata/converter/ListConverter.kt
JcMinarro: Because we had that into the project:
https://github.com/GetStream/stream-chat-android-livedata/pull/49/files#diff-1e70daafb475c0ce3fef7d2728279182

ogkuzmin (approved):

PR #52, opened 2020-09-21
Fix/incorrect users mapping in channel

ogkuzmin:
Fix incorrect mapping of users for channel.


tschellenbach:
this indeed fixes the crash

JcMinarro (approved):

PR #57, opened 2020-09-24
Draft, raise error when calling .execute on main thread

tschellenbach:
Throw an exception when someone calls .execute in the main thread for use cases that perform IO. There are probably more ways to achieve the same goal. Either way I think it's better to fail and not allow this. 

Don't merge, its a draft to discuss if we should raise an error.

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/utils/CallImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/utils/CallImpl.kt
@@ -1,5 +1,6 @@
 package io.getstream.chat.android.livedata.utils
 
+import android.os.Looper
 import androidx.annotation.UiThread
 import androidx.annotation.WorkerThread
 import io.getstream.chat.android.client.utils.Result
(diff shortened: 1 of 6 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/utils/CallImpl.kt
```diff
@@ -30,8 +32,11 @@ class CallImpl2<T>(var runnable: suspend () -> Result<T>, var scope: CoroutineSc
     }
 
     override fun execute(): Result<T> {
-        val result = runBlocking(scope.coroutineContext) { runnable() }
-        return result
+        val isMainThread = Looper.getMainLooper().thread == Thread.currentThread()
```
ogkuzmin, line 35: What if we mark an interface method with `@Throw` annotation?

livedata/src/main/java/io/getstream/chat/android/livedata/utils/CallImpl.kt
```diff
@@ -21,7 +23,7 @@ interface Call2<T> {
     suspend fun invoke(): Result<T>
 }
 
-class CallImpl2<T>(var runnable: suspend () -> Result<T>, var scope: CoroutineScope = GlobalScope) :
+class CallImpl2<T>(var runnable: suspend () -> Result<T>, var scope: CoroutineScope = GlobalScope, var mainThreadAllowed: Boolean = false) :
```
ogkuzmin, line 26: Do we really need var instead of val?

PR #59, opened 2020-09-29
Fix/edit message

tschellenbach:
- enable replayEventsForActiveChannels
- fix a bug with memberLimit vs messageLimit
- settings.gradle update for test cycle
- fix reaction crash
- sendMessage copy the message to prevent issues with mutability & diffutils. This was causing the message failed to send to not show accurately
- fix a bug where network errors (temporary ones) are detected as permanent errors

**Edit Message now works well offline**

- when making multiple edit message requests while offline, only run the last one and cancel prior jobs
- copy the message to prevent issues with mutating the object
- fix an issue where incorrect message.updatedAt dates returned by the backend could prevent an object from being updated


livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -988,42 +997,62 @@ class ChannelControllerImpl(
...
+        // set message.updated at if it's null or older than now (prevents issues with incorrect clocks)
+        if (editedMessage.updatedAt == null) {
+            editedMessage.updatedAt = now
+        }
+        editedMessage.updatedAt?.let {
+            if (it.before(now)) {
+                editedMessage.updatedAt = now
+            }
+        }
+
```
tschellenbach, line 1012: not happy with this verbosity, what's the shortest way to set `editMessage.updatedAt` to now when it's null or before `now`?

livedata/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -85,7 +85,7 @@ fun ChatError.isPermanent(): Boolean {
     var isPermanent = true
     if (this is ChatNetworkError) {
         val networkError: ChatNetworkError = this
-        if (networkError.statusCode == 429) {
+        if (networkError.statusCode == 429 || networkError.statusCode == -1) {
```
JcMinarro, line 88: Could we extract those magic number?
BTW, the comment that precedes the code is not up-to-date.
This kind of comments should be removed, they aren't mantained with the new changes are done on the code and could cause confusion.
Instead of that, we need to write some logic that explain the process.
Instead of check with some magic number, they could be into a constant that indicate which status code are permanent error, and check them

livedata/src/main/java/io/getstream/chat/android/livedata/utils/DefaultRetryPolicy.kt
```diff
@@ -9,7 +9,7 @@ class DefaultRetryPolicy : RetryPolicy {
         attempt: Int,
         error: ChatError
     ): Boolean {
-        return attempt < 5
+        return attempt < 3
```
JcMinarro, line 12: Could we extract this magic number?

JcMinarro (commented):
This PR mixes multiple topics, Could we split it into smaller PRs that doesn't mix unrelated topics

samiuelson (changes requested):

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ReactionEntity.kt
```diff
@@ -46,7 +46,7 @@ data class ReactionEntity(@PrimaryKey var messageId: String, var userId: String,
     var syncStatus: SyncStatus = SyncStatus.COMPLETED
 
     /** create a reactionEntity from a reaction object */
-    constructor(r: Reaction) : this(r.messageId, r.user!!.id, r.type) {
+    constructor(r: Reaction) : this(r.messageId, r.fetchUserId(), r.type) {
```
samiuelson, line 49: seems `Reaction.fetchUserId()` function is not there in LLC. Is there a PR related to this change somewhere?

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
samiuelson: added code suggestion above

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -988,42 +997,62 @@ class ChannelControllerImpl(
...
+        // set message.updated at if it's null or older than now (prevents issues with incorrect clocks)
+        if (editedMessage.updatedAt == null) {
+            editedMessage.updatedAt = now
+        }
+        editedMessage.updatedAt?.let {
+            if (it.before(now)) {
+                editedMessage.updatedAt = now
+            }
+        }
```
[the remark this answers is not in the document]
tschellenbach: clean syntax, editor tells me it needs a !!
bit weird since the if statement already checks for null 

```
editedMessage.apply {
            val now = Date()
            if (updatedAt == null || updatedAt!!.before(now)) {
                updatedAt = now
            }
        }
```

livedata/src/main/java/io/getstream/chat/android/livedata/utils/DefaultRetryPolicy.kt
tschellenbach: nothing magic about it, just the default is 3 retries.

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ReactionEntity.kt
tschellenbach: its on the develop branch atm.

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -484,7 +484,7 @@ class ChatDomainImpl private constructor(
         }
 
     private fun queryEvents(cids: List<String>): List<ChatEvent> {
-        val response = client.getSyncHistory(cids, syncState?.lastSyncedAt ?: NEVER).execute()
+        val response = client.getSyncHistory(cids, syncState?.lastSyncedAt ?: Date()).execute()
```
samiuelson, line 487: If l`astSyncAt` is null should we pass the current timestamp? This is a value that's narrowing the results, right?

samiuelson (approved):

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
tschellenbach: Never doesn't work since you hit the limit of how far back you can go and the request is ignored. @tbarbugli what do you think is the right default?

PR #61, opened 2020-09-30
Notifications improvements

samiuelson:
* Adds support for additional values send through firebase cloud message: channel_name, message_text
* Minor code clean-up

Should be merged after LLC 1.16.2 is released

### DIFF
--- a/livedata/build.gradle
+++ b/livedata/build.gradle
@@ -68,7 +68,7 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
 
-    api "com.github.getstream:stream-chat-android-client:1.16.1"
+    api "com.github.getstream:stream-chat-android-client:1.16.4-SNAPSHOT"
 
     def room_version = "2.2.5"
     def work_version = "2.4.0"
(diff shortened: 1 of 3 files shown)


livedata/build.gradle
```diff
@@ -139,7 +139,7 @@ afterEvaluate {
                 // Adds sources as separate jars.
                 artifact androidSourcesJar
                 // Custom attributes of the publication
-                groupId = project.group
+                groupId = 'com.github.getstream'
```
JcMinarro, line 142: Why are we changing it?

samiuelson: If `publishToMavenLocal` is done with 'com.github.GetStream' group there is a metadata error while fetching the artifact in the target project. Or is it only locally on my end?

JcMinarro: I didn't have this problem before.
BTW, Can't you change the `project.group` value?

samiuelson: Not sure how jitpack will react. I will probably get stuck finding repo on github

JcMinarro (approved):

PR #62, opened 2020-09-30
Fix/send message with attachments - Draft

tschellenbach:
- deprecate SendMessageWithAttachments
- move the logic of uploading files to SendMessage
- the idea here is to have insert message into local storage, upload, rerender the message. (the previous logic waited for the upload to finish before writing the message/updating the UI)
- this will also retry message uploads upon calling connection recovered
- Store image upload failure status

Video: https://www.loom.com/share/245b3a08b47b425cbcfbef71a098048f

Later:

- WorkManager based file upload
- Show local image in the UI with progress indicator (after message list refactor is done)
- Progress indicator (after message list refactor is done)


samiuelson:
Amazing video demo 👍🏼

PR #63, opened 2020-10-02
Request for members by default when querying  channels

JcMinarro:
Members weren't requested by default when asking for channels, so the members list obtained was empty. It is an issue for DM Channels where the picture rendered on the channels list is the picture of the other user. Happend the same with the Channel Name. 

Fixes https://github.com/GetStream/stream-chat-android/issues/644

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
@@ -24,6 +24,10 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 
+private const val MESSAGE_LIMIT = 10
+private const val MEMBER_LIMIT = 30
+private const val INITIAL_CHANNEL_OFFSET = 0
+private const val CHANNEL_LIMIT = 30
 class QueryChannelsControllerImpl(
     override val filter: FilterObject,
     override val sort: QuerySort,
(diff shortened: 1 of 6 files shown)


adasiewiczr (approved):

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -24,6 +24,10 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 
+private const val MESSAGE_LIMIT = 10
```
ogkuzmin, line 27: Nit: It's a matter of taste, but such constant looks better in the companion object of such class

ogkuzmin (approved):

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
zsmb13: Same, having them nested in the companion object (or just another named object) inside the class is more natural for me, makes the constants belong specifically to the class.

(And it makes the icon for the file looks nicer in the Project pane of the IDE)

JcMinarro: When this constant is used outside of the class, I see the point to use the companion object and have the code more accessible/ordered, but for private constants that are used only inside of this file, I don't see any benefit, it looks more like "Java-Style"

zsmb13: In theory, we could have multiple classes in the same file. If only one of those uses a given constant, I'd have place the constants inside the class, to have it exist in the smallest scope possible.

Placing it inside the class, even while there's only one in the file, puts it in a narrower scope than having it at the file level, even if private.

tschellenbach:
@tbarbugli don't we return members by default on the API side of things?

PR #64, opened 2020-10-05
Cas 212/refac user entity

ogkuzmin:
Refactoring DB work with UserEntity
1) Made UserEntity immutable
2) Made UserEntity internal and hide from clients
3) Made ChatDatabase internal and hide from public API
4) Change tests from integration to unit ones


livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -46,9 +46,7 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.delay
 import java.lang.Thread.sleep
-import java.util.Date
-import java.util.InputMismatchException
-import java.util.UUID
+import java.util.*
```
tschellenbach, line 49: my editor has the same issue

JcMinarro: It can be configured into AS Setting "Editor > Code Style > Kotlin" and only allow "Use single name import" and remove `java.util` from "package to use import with '*'" section

livedata/src/main/java/io/getstream/chat/android/livedata/repository/UserRepository.kt
```diff
@@ -5,81 +5,89 @@ import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.livedata.dao.UserDao
 import io.getstream.chat.android.livedata.entity.UserEntity
 
-class UserRepository(
-    var userDao: UserDao,
-    var cacheSize: Int = 100,
-    var currentUser: User?
+internal class UserRepository(
+    private val userDao: UserDao,
+    private val currentUser: User?,
```
JcMinarro, line 10: Does it make sense to have a "null `currentUser`"?
Can't we delay the `UserRepository` initialization until we have a user logged in?

livedata/src/main/java/io/getstream/chat/android/livedata/repository/UserRepository.kt
```diff
@@ -5,81 +5,89 @@ import io.getstream.chat.android.client.models.User
...
-        insert(listOf(userEntity))
+        userDao.insert(toEntity(user))
     }
 
     suspend fun insertMe(user: User) {
-        val userEntity = UserEntity(user)
-        userEntity.originalId = user.id
-        userEntity.id = "me"
-        insert(listOf(userEntity))
+        val userEntity = toEntity(user).copy(originalId = user.id, id = ME_ID)
```
JcMinarro, line 33: Could we move it into `toEntity()` method?
We can assign `originalId` as `user.id` to all entities, and only edit the `id` when it is the current user.
It would simplify the `toModel()` method

JcMinarro (commented):
Great to have "User Entities Stuff" only on "DB Layer"

tschellenbach:
All of the entity objects should be immutable. Good refactor. Just make sure that linting and tests are happy.

livedata/src/main/java/io/getstream/chat/android/livedata/repository/UserRepository.kt
ogkuzmin: Yes, it's definitely possible, since RepositoryHelper already contains user in its' constructor.

livedata/src/main/java/io/getstream/chat/android/livedata/repository/UserRepository.kt
ogkuzmin: Yes, good idea. Implemented it

JcMinarro (dismissed):

JcMinarro (approved):

PR #70, opened 2020-10-10
Fix/performance

tschellenbach:
**Indexes**

* Message/Channel/Reaction now have an index on syncStatus
* Message has a new index on [message.cid,message.createdAt]

**Last Message**

For our sample app (and basically any social chat type of app) it's common to show the last message in the channel list. This PR denormalizes the last message on the channel object which shaves off ~200ms for opening the channel list.

Also updated the default on messageLimit for queryChannels to be 1 (which is a more sensible default given most chat interfaces)

**Start API calls earlier**

Changed when the API calls for query channel and query channels start. Since offline storage has quite a bit of work to do it makes sense to start the API call at the same time as we query the DB.

**Querying messages**

We now run the queries for getting messages for a certain channel in parallel. I didn't notice a performance improvement from this, but not entirely sure if that's device/emulator specific. 

**Bug fixes**

Offline storage would show the oldest instead of the newest messages by default. Fixed that.


livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -607,7 +608,7 @@ class ChatDomainImpl private constructor(
         // store the messages
         repos.messages.insertMessages(messages)
 
-        logger.logI("stored ${channelsResponse.size} channels, ${configs.size} configs, ${users.size} users and ${messages.size} messages")
+        logger.logI("storeStateForChannels stored ${channelsResponse.size} channels, ${configs.size} configs, ${users.size} users and ${messages.size} messages")
```
leandroBorgesFerreira, line 611: Is this some left over code?

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -173,22 +173,38 @@ class EventHandlerImpl(
...
+                        it.addMessage(MessageEntity(event.message))
+                        channels[it.cid] = it
+                    }
                 }
                 is NotificationMessageNewEvent -> {
                     event.message.cid = event.cid
                     event.totalUnreadCount?.let { domainImpl.setTotalUnreadCount(it) }
                     users.putAll(event.message.users().associateBy(User::id))
                     messages[event.message.id] = MessageEntity(event.message)
+                    channelMap[event.cid]?.let {
```
leandroBorgesFerreira, line 204: You could perhaps extract this code to a function, it is being repeated 4 times.

livedata/src/main/java/io/getstream/chat/android/livedata/request/AnyChannelPaginationRequest.kt
```diff
@@ -30,11 +30,19 @@ internal class AnyChannelPaginationRequest(var messageLimit: Int = 30) {
         return messageFilterDirection == null
     }
 
+    fun isRequestingMoreThanLastMessage(): Boolean {
+        return (isFirstPage() && messageLimit > 1) || (isNotFirstPage() && messageLimit > 0)
+    }
```
zsmb13, line 35: The two new functions added to this class are good candidates for extension functions instead of members, since they just give us "views" over the publicly exposed data of the class.

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
```diff
@@ -77,7 +83,12 @@ data class ChannelEntity(var type: String, var channelId: String) {
         for (r in c.read) {
             reads[r.getUserId()] = ChannelUserReadEntity(r)
         }
-        lastMessageAt = c.lastMessageAt
+        if (c.messages.isNotEmpty()) {
+            lastMessage = MessageEntity(c.messages.last())
+            lastMessage?.let {
+                lastMessageAt = it.createdAt
+            }
+        }
```
zsmb13, line 91:
```suggestion
        c.messages.lastOrNull()?.let { message ->
            lastMessage = MessageEntity(message)
            lastMessageAt = message.createdAt
        }
```

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
zsmb13: The same for some of the previous lines in the different branches here

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -639,11 +640,16 @@ class ChatDomainImpl private constructor(
...
+        val channelMessagesMap = if (pagination.isRequestingMoreThanLastMessage()) {
             // with postgres this could be optimized into a single query instead of N, not sure about sqlite on android
             // sqlite has window functions: https://sqlite.org/windowfunctions.html
             // but android runs a very dated version: https://developer.android.com/reference/android/database/sqlite/package-summary
-            channelEntities.map { it.cid to repos.messages.selectMessagesForChannel(it.cid, pagination) }.toMap()
+
+            // async to run queries in parallel
+            channelEntities.map { scope.async { it.cid to repos.messages.selectMessagesForChannel(it.cid, pagination) } }.awaitAll().toMap()
+            // sync version:
+            // channelEntities.map { it.cid to repos.messages.selectMessagesForChannel(it.cid, pagination) }.toMap()
```
zsmb13, line 652: Let's remove the old code, it's in VCS anyway, if we need it again sometime

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
zsmb13: The same for some of the previous lines in the different branches here

tschellenbach:
Alright, I only work a few hours in my spare time on Android. I think I can look at this on Wednesday, but if you need to merge this earlier I won't be insulted if you edit it directly.

zsmb13:
I'll pick up those fixes tomorrow morning then 👍

livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
zsmb13: It's the method name, indicating _what_ stored those channels. Think "the `storeStateForChannels` method stored..."

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
zsmb13: Named this `addMessageData`, but happy to hear more meaningful suggestions

tschellenbach:
updated the codebase with the recent changes from develop.

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
```diff
@@ -105,9 +118,17 @@ data class ChannelEntity(var type: String, var channelId: String) {
...
-    /** updates last message and lastmessagedate on this channel entity */
-    fun updateLastMessageDate(messageEntity: MessageEntity) {
-        lastMessageAt = max(lastMessageAt, messageEntity.createdAt ?: messageEntity.createdLocallyAt)
+    /** updates last message and lastMessageAt on this channel entity */
+    fun addMessage(messageEntity: MessageEntity) {
+        val createdAt = messageEntity.createdAt ?: messageEntity.createdLocallyAt
+        val messageEntityCreatedAt = checkNotNull(createdAt) { "created at cant be null, be sure to set message.createdAt" }
+
+        val updateNeeded = messageEntity.id == lastMessage?.id
+        val newLastMessage = lastMessageAt == null || messageEntityCreatedAt.after(messageEntityCreatedAt)
```
zsmb13, line 127: I think this is the comparison that we're meant to have here? Not sure how we ended up with the current version of the line.

```suggestion
        val newLastMessage = lastMessageAt == null || messageEntityCreatedAt.after(lastMessageAt)
```

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -382,32 +382,42 @@ class ChannelControllerImpl(
...
+        return if (queryOnlineJob != null) {
+            val response = queryOnlineJob.await()
+            if (response.isSuccess) {
+                updateLiveDataFromChannel(response.data())
+            }
+            response
         } else {
             // if we are not offline we mark it as needing recovery
             recoveryNeeded = true
+            val channel = channelEntityPair?.channel
```
JcMinarro, line 404: If `channelEntityPair` is null, we are returning a `Result(null, null)` that doesn't make sense

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
```diff
@@ -105,9 +118,17 @@ data class ChannelEntity(var type: String, var channelId: String) {
         return c
     }
 
-    /** updates last message and lastmessagedate on this channel entity */
-    fun updateLastMessageDate(messageEntity: MessageEntity) {
-        lastMessageAt = max(lastMessageAt, messageEntity.createdAt ?: messageEntity.createdLocallyAt)
+    /** updates last message and lastMessageAt on this channel entity */
+    fun addMessage(messageEntity: MessageEntity) {
```
JcMinarro, line 122: This method name is a bit confusing. It is not adding any message to any part

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
JcMinarro: Yes, it is the needed comparison

tschellenbach: Good spot

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
tschellenbach: renamed it to updateLastMessage

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
tschellenbach: the offline result has 3 possible state. unknown (null), empty or an actual channel. I don't think the behaviour changed during this refactor.

tschellenbach: confirmed, this didn't change: https://github.com/GetStream/stream-chat-android-livedata/blob/master/livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt#L383

zsmb13 (approved):
Good to go as far as I'm concerned 👍

PR #73, opened 2020-10-12
Handle not found members into our DB

JcMinarro:
### Description

By some reason we are not storing all members of a channels into our DB, and when we ask for members of a channel it is crashing.
This part is going to be refactored by @ogkuzmin into another task, but for now we can avoid this NPE by omitting this member

### Checklist

- [ ] Changelog updated with client-facing changes
- [ ] New code is covered by unit tests
- [x] Reviewers added

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/entity/MemberEntity.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/entity/MemberEntity.kt
@@ -1,5 +1,6 @@
 package io.getstream.chat.android.livedata.entity
 
+import io.getstream.chat.android.client.logger.ChatLogger
 import io.getstream.chat.android.client.models.Member
 import io.getstream.chat.android.client.models.User
 import java.util.Date
@@ -36,15 +37,8 @@ data class MemberEntity(var userId: String) {
     }
 
     /** converts a member entity into a member */
-    fun toMember(userMap: Map<String, User>): Member {
-        val user = userMap[userId] ?: error("userMap is missing the user for this member")
-        val r = Member(user, role)
-        r.createdAt = createdAt
-        r.updatedAt = updatedAt
-        r.isInvited = isInvited
-        r.inviteAcceptedAt = inviteAcceptedAt
-        r.inviteRejectedAt = inviteRejectedAt
-
-        return r
-    }
+    fun toMember(userMap: Map<String, User>): Member? =
+        userMap[userId]?.let {
+            Member(it, role, createdAt, updatedAt, isInvited, inviteAcceptedAt, inviteRejectedAt)
+        } ?: null.also { ChatLogger.get("MemberEntity").logE("userMap is missing the user with id='$userId` needed to create this member") }
 }
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/entity/MemberEntity.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/entity/MemberEntity.kt
@@ -37,8 +37,13 @@ data class MemberEntity(var userId: String) {
     }
 
     /** converts a member entity into a member */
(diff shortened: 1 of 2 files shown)


tschellenbach:
there are 2 flows where we receive new members

- storeStateForChannels
- events

ogkuzmin (dismissed):

tschellenbach:
storeStateForChannels looks correct, seems related to an event

JcMinarro:
> there are 2 flows where we receive new members
> 
> * storeStateForChannels
> * events
Member are paginated now, maybe it is the cause because an user is not found. 
BTW, some changes are been doing on this part of the code to improve our persistence layer behavior, on this refactor process we will cover it

tschellenbach:
there are some events that update the channel directly without also storing the channel users.

tschellenbach:
good solution

adasiewiczr (approved):

livedata/src/main/java/io/getstream/chat/android/livedata/entity/MemberEntity.kt
```diff
@@ -36,15 +37,8 @@ data class MemberEntity(var userId: String) {
...
-        r.isInvited = isInvited
-        r.inviteAcceptedAt = inviteAcceptedAt
-        r.inviteRejectedAt = inviteRejectedAt
-
-        return r
-    }
+    fun toMember(userMap: Map<String, User>): Member? =
+        userMap[userId]?.let {
+            Member(it, role, createdAt, updatedAt, isInvited, inviteAcceptedAt, inviteRejectedAt)
+        } ?: null.also { ChatLogger.get("MemberEntity").logE("userMap is missing the user with id='$userId` needed to create this member") }
```
zsmb13, line 43: I know it's boring code, but wouldn't this be more readable than `x?.let { } ?: null.also { }`? The `null.also` part is especially making me uncomfortable 😅 

```suggestion
    fun toMember(userMap: Map<String, User>): Member? {
        val user = userMap[userId]
        if (user == null) {
            ChatLogger.get("MemberEntity")
                .logE("userMap is missing the user with id='$userId` needed to create this member")
            return null
        }
        return Member(user, role, createdAt, updatedAt, isInvited, inviteAcceptedAt, inviteRejectedAt)
    }
```

JcMinarro: I would prefer that Kotlin provides some kind of sideefect extension for Elvis operator, but it is only possible for "sucess path", and the propose code seems "so old java style"..., but well, we can change it

zsmb13: It _is_ "Java style", but it's also very easily parseable, anyone looking at it will know what it does immediately, and not have to wonder about what `also` is called on (just `null`, or the entire Elvis expression), or have to think through what happens exactly when something along the way in the chain is `null`.

You're basically unrolling the Kotlin style code into this in your head anyway, which is extra processing 😄

tschellenbach: 👏 simple is better. It's ok to spend a few extra lines for readability.

(coming from someone who used to love abusing python metaclasses for side effects. Nowadays I really like languages like Go which basically don't allow you to get creative. )

That being said, Kotlin can have some issues with nullability and if statements. It doesn't always understand that it's in a clause where the value can't be null.

zsmb13: In the cases you're thinking of, the compiler is actually (at least technically) correct. It doesn't give you smart casts in `if` checks if you have a mutable, non-local variable, because it would be unsafe in a multithreaded environment. I've written this down in detail [here](https://stackoverflow.com/a/51094692/4465208).

PR #75, opened 2020-10-14
Improve sync data validation in ChatDomain.Builder

samiuelson:
### Description
Improve sync data validation in ChatDomain.Builder

### Checklist
- [x] Changelog updated with client-facing changes
- [x] New code is covered by unit tests
- [x] Reviewers added

### DIFF
--- a/livedata/build.gradle
+++ b/livedata/build.gradle
@@ -138,7 +138,7 @@ afterEvaluate {
                 // Custom attributes of the publication
                 groupId = project.group
                 artifactId = 'stream-chat-android-livedata'
-                version = '0.8.6-SNAPSHOT'
+                version = '0.8.7-SNAPSHOT'
             }
         }
     }
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -96,6 +96,7 @@ interface ChatDomain {
 
         private var userPresence: Boolean = false
         private var offlineEnabled: Boolean = true
+        private var backgroundSyncEnabled: Boolean = true
         private var recoveryEnabled: Boolean = true
         private var backgroundSyncConfig: BackgroundSyncConfig = BackgroundSyncConfig.UNAVAILABLE
         private var notificationConfig: NotificationConfig = NotificationConfigUnavailable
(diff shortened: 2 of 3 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -108,9 +108,11 @@ interface ChatDomain {
 
         fun backgroundSyncEnabled(apiKey: String, userToken: String): Builder {
             // TODO: Consider exposing apiKey and userToken by ChatClient to make this function more friendly
-            this.backgroundSyncConfig = BackgroundSyncConfig(apiKey, user.id, userToken)
             if (apiKey.isEmpty() || user.id.isEmpty() || userToken.isEmpty()) {
-                throw IllegalArgumentException("ChatDomain.Builder::backgroundSyncEnabled. apiKey and user.id, and userToken must not be empty.")
+                this.backgroundSyncConfig = BackgroundSyncConfig.UNAVAILABLE
+                throw IllegalArgumentException("ChatDomain.Builder::backgroundSyncEnabled. apiKey, user.id, and userToken must not be empty.")
```
zsmb13, line 113: Do we really expect clients to catch the exception and then proceed to use the same `Builder`? If they do something invalid with it, I think we should mark that instance as broken and never let it `build`. That's something they should fix in their client code, they shouldn't get away with it 😄

samiuelson: In case Builder instance is "broken" would you throw an exception inside `build()` instead? Or do you have some other idea of preventing instantiation of ChatDomain 🤔

zsmb13: Yup, that's what I was thinking. Somehow invalidate it to prevent anyone from using it, if they've ever made an invalid call to it.

adasiewiczr (approved):
LGTM

livedata/build.gradle
```diff
@@ -139,7 +139,7 @@ afterEvaluate {
                 // Custom attributes of the publication
                 groupId = project.group
                 artifactId = 'stream-chat-android-livedata'
-                version = '0.8.6-SNAPSHOT'
+                version = '0.8.7-SNAPSHOT'
```
JcMinarro, line 142: This version wouldn't need to be updated. Last released version is 0.8.5, so `0.8.6-SNAPSHOT` should be enough.
It is updated everytime that a new version is released

zsmb13 (approved):

PR #76, opened 2020-10-14
Fix/event handler npe

JcMinarro:
### Description
For some reason backend is not sending us the user instance into some events that they should and we are not able to identify which event type is. Gson, because it is using reflection, inject a null instance into property `user` that doesn't allow null values. This is a workaround, while we identify which event type is, that omit null values without break our public API

### Checklist

- [x] Changelog updated with client-facing changes
- [ ] New code is covered by unit tests
- [x] Reviewers added

Fix https://github.com/GetStream/stream-chat-android/issues/685

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
@@ -211,18 +211,6 @@ class EventHandlerImpl(
                         hidden = false
                     }
                 }
-                is ChannelUserBannedEvent -> {
-                    users[event.user.id] = event.user
-                }
-                is GlobalUserBannedEvent -> {
-                    users[event.user.id] = event.user.apply { banned = true }
-                }
-                is ChannelUserUnbannedEvent -> {
-                    users[event.user.id] = event.user
-                }
-                is GlobalUserUnbannedEvent -> {
-                    users[event.user.id] = event.user.apply { banned = false }
-                }
                 is NotificationMutesUpdatedEvent -> {
                     domainImpl.updateCurrentUser(event.me)
                 }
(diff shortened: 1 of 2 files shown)


ogkuzmin (dismissed):

samiuelson (approved):

samiuelson (dismissed):

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -95,7 +95,12 @@ class EventHandlerImpl(
         val messagesToFetch = mutableSetOf<String>()
 
         events.filterIsInstance<CidEvent>().onEach { channelsToFetch += it.cid }
-        users += events.filterIsInstance<UserEvent>().map(UserEvent::user).associateBy(User::id)
+        // For some reason backend is not sending us the user instance into some events that they should
+        // and we are not able to identify which event type is. Gson, because it is using reflection,
+        // inject a null instance into property `user` that doesn't allow null values.
+        // This is a workaround, while we identify which event type is, that omit null values without
+        // break our public API
+        users += events.filterIsInstance<UserEvent>().mapNotNull { it.user as User? }.associateBy(User::id)
```
zsmb13, line 103: Hurts that we have to do a meaningless cast like this, that's logically incorrect given our types, but since our types are incorrect... Anyway, let's at least suppress the warning, so that we don't accidentally remove it 😄 

```suggestion
        @Suppress("USELESS_CAST")
        users += events.filterIsInstance<UserEvent>().mapNotNull { it.user as User? }.associateBy(User::id)
```

zsmb13 (dismissed):

livedata/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
JcMinarro: Yes, I am not happy with this implementation neither. Hope to remove it soon

zsmb13 (approved):

PR #80, opened 2020-10-16
first notes on sort problem

### LINKED ISSUE #21: Incorrect work of QuerySort

tschellenbach:
Depend on https://github.com/GetStream/stream-chat-android-client/pull/110

Fix #21 

the sort changes between online and offline query. this is caused by the offline sort not working. 

see this video for clarification: https://www.loom.com/share/d956c3cb4353480ba9c321a9d026de0d

I investigated it a little bit, few issues:

1. - the backend uses field names like created_at which on the channel object are called createdAt, these don't match
2. - the default ordering is last_updated which is actually a composite field based on max(createdAt, lastMessageAt)
3. - you can receive new channels via an event. the current sort implementation is not applied after receiving an event. it's only used for the offline version of the channels.

I added a failing test for the first 2 issues.

1. - i added a fix for this. number 3 is also easy to solve by changing that Transform where I added a comment.

supporting last_updated is a bit harder, any ideas?

- maybe an extension function on the Channel object?

### DIFF
--- a/livedata/src/test/java/io/getstream/chat/android/livedata/PaginationTest.kt
+++ b/livedata/src/test/java/io/getstream/chat/android/livedata/PaginationTest.kt
@@ -1,11 +1,11 @@
 package io.getstream.chat.android.livedata
 
-import com.google.common.truth.Truth
 import io.getstream.chat.android.client.api.models.QuerySort
 import io.getstream.chat.android.livedata.entity.ChannelEntityPair
 import io.getstream.chat.android.livedata.extensions.applyPagination
 import io.getstream.chat.android.livedata.request.AnyChannelPaginationRequest
 import io.getstream.chat.android.livedata.utils.calendar
+import org.amshove.kluent.`should be equal to`
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.Arguments
 import org.junit.jupiter.params.provider.MethodSource
@@ -20,8 +20,7 @@ internal class PaginationTest {
         expectedList: List<ChannelEntityPair>
     ) {
         // show an easy to use diff between the two results
-        val result = inputList.applyPagination(pagination).map { it.channel.cid }
-        Truth.assertThat(result).isEqualTo(expectedList.map { it.channel.cid })
+        inputList.applyPagination(pagination) `should be equal to` expectedList
     }
 
     companion object {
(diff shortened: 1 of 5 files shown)


livedata/src/test/java/io/getstream/chat/android/livedata/PaginationTest.kt
```diff
@@ -18,7 +19,9 @@ class PaginationTest {
         pagination: AnyChannelPaginationRequest,
         expectedList: List<ChannelEntityPair>
     ) {
-        inputList.applyPagination(pagination) `should be equal to` expectedList
+        // show an easy to use diff between the two results
```
JcMinarro, line 22: Check only the `cid` doesn't cover all cases. It is assuming our `cid` will be unique, but it is something that we can't guarantee with our own models.
Apart of that, test should guarantee the full list content is the one we expect

livedata/src/test/java/io/getstream/chat/android/livedata/PaginationTest.kt
```diff
@@ -162,6 +165,54 @@ class PaginationTest {
                     },
                     it
                 )
+            },
+            // last_updated is a computed field based on max(createdAt, lastMessageAt)
```
JcMinarro, line 169: `last_updated` is not a field on `Channel` Wee need to add it

livedata/src/test/java/io/getstream/chat/android/livedata/PaginationTest.kt
tschellenbach: `cid` is a unique identifier for channels

livedata/src/test/java/io/getstream/chat/android/livedata/PaginationTest.kt
tschellenbach: what do you think is best? want to add a property on the low level client that computes the max(createdAt, lastMessageAt)?

tschellenbach: i saw the other PR, nice

JcMinarro (approved):

PR #83, opened 2020-10-16
debugging the offline update issue

tschellenbach:
Investigation into https://github.com/GetStream/stream-chat-android/issues/698

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
@@ -477,8 +477,13 @@ internal class ChannelControllerImpl(
 
         val messageEntity = MessageEntity(newMessage)
 
-        // Update livedata
+        // Update livedata in channel controller
         upsertMessage(newMessage)
+        // TODO: an event broadcasting feature for LOCAL/offline events on the LLC would be a cleaner approach
+        // Update livedata for currently running queries
+        for (query in domainImpl.getActiveQueries()) {
+            query.broadcastChannelUpdate(cid)
+        }
 
         // we insert early to ensure we don't lose messages
         domainImpl.repos.messages.insertMessage(newMessage)
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
@@ -103,6 +103,12 @@ internal class QueryChannelsControllerImpl(
         }
     }
 
+    internal fun broadcastChannelUpdate(cid: String) {
+        val channel = domainImpl.channel(cid).toChannel()
+        logger.logI("channel ${channel.cid} manually broadcasting update, last message text is ${channel.messages.lastOrNull()?.text}")
+        updateChannel(channel)
+    }
+
     private fun updateChannel(channel: Channel) {
         _channels.postValue((_channels.value ?: mapOf()) + mapOf(channel.cid to channel))
     }


carterhudson (dismissed):

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -469,6 +469,7 @@ internal class ChannelControllerImpl(
         }
 
         newMessage.user = domainImpl.currentUser
+        newMessage.type = "regular"
```
JcMinarro, line 472: Is it always `regular` type? 
Could we use a constant value instead of this hardcoded string?

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -89,20 +89,29 @@ internal class QueryChannelsControllerImpl(
 
     fun handleEvent(event: ChatEvent) {
         if (event is NotificationAddedToChannelEvent) {
+            // this is the only event that adds channels to the query
             addChannelIfFilterMatches(event.channel)
-        }
-        if (event is CidEvent) {
+        } else if (event is CidEvent) {
```
JcMinarro, line 94: `NotificationAddedToChannelEvent` is a `CidEvent` as well, why do we skip it?

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -50,9 +51,8 @@ internal class QueryChannelsControllerImpl(
     override val endOfChannels: LiveData<Boolean> = _endOfChannels
 
     private val _channels = MutableLiveData<Map<String, Channel>>()
-    // Ensure we don't lose the sort in the channel
-    // TODO: The live list of channels doesn't actually maintain the specified query sort
-    override var channels: LiveData<List<Channel>> = Transformations.map(_channels) { cMap -> queryEntity.channelCids.mapNotNull { cMap[it] } }
+    // Keep the channel list locally sorted
+    override var channels: LiveData<List<Channel>> = Transformations.map(_channels) { cMap -> cMap.values.sortedWith(sort.channelComparator) }
```
JcMinarro, line 55: Have this change into another PR would be easier and faster to be reviewed

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
tschellenbach: Regular is a good default, ideally we would check for slash commands and do regular or ephemeral based on that.

Type should be an enum/sealed class in my opinion

samiuelson (approved):

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
tschellenbach: going to clarify this, it's confusing

tschellenbach:
Don't merge yet, i want to process some of JCs feedback

JcMinarro:
> Don't merge yet, i want to process some of JCs feedback

You can mark it as "draft", then it is not able to be merged until it comeback to be ready to review.
Let me mark it as draft for now

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -519,6 +525,9 @@ internal class ChannelControllerImpl(
                 upsertMessage(processedMessage)
                 Result(processedMessage, null)
             } else {
+
+                logger.logE("Failed to send message with id ${newMessage.id} and text ${newMessage.text}", result.error())
```
tschellenbach, line 529: @JcMinarro is this the best way to log an error? I noticed on the sample app that messages sometimes fail to send. It only happens on my real device though, can't replicate it on the emulator. So how do we log an error and have it show up in crashlytics/firebase?

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
tschellenbach: clarified which method calls update the results vs which ones update the channel state. that was confusing.

livedata/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
tschellenbach: I like to test the sample app thoroughly before merging a PR, bit too much overhead to split it up.

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: With the current implementation we have we are not sending any log to Firebase, only crashes

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
```diff
@@ -33,7 +33,7 @@ internal data class ChannelEntity(var type: String, var channelId: String) {
     var frozen: Boolean = false
 
     /** if the channel is hidden (new messages will cause to reappear) */
-    var hidden: Boolean = false
+    var hidden: Boolean? = null
```
JcMinarro, line 36: Why do we need it as nullable property?

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -469,6 +469,8 @@ internal class ChannelControllerImpl(
         }
 
         newMessage.user = domainImpl.currentUser
+        // TODO: type should be a sealed/class or enum at the client level
+        newMessage.type = if (newMessage.text.startsWith("/")) { "ephemeral" } else { "regular" }
```
JcMinarro, line 473: Could we extract those hardcoded values out of the code?
Ideally, `type` should be an enum, but at least we should keep them into some constant instead of having hardcoded value around the code

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
tschellenbach: otherwise the difference between null and false triggers the channel diff callback to believe something has changed.

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -465,6 +465,8 @@ internal class ChannelControllerImpl(
...
```
samiuelson, line 469: Is it safe to assume that message is ephemeral or not by `tartsWith("/")` condition? 
E.g. What if the backend decides that message with text like `"/////"` is of `regular` type instead? Isn't it going to break something?

tschellenbach: no it will just update once the backend tells us what it really is. it's just a sensible default so that we actually show the message in the channel list (it filters on ==regular)

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
tschellenbach: we have that in the UI layer already, lets just fix it properly

livedata/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
samiuelson: So the filter will not allow showing messages starting with "/", right? I think it might cause some confusion in the future. Anyway, it's rare to start a chat message with this char 😄

tschellenbach: so if you send a message that is something like `/hi there`
it won't show up in the channel list while offline. as soon as the MessageEvent is received it will update to the correct state. (or actually probably raise an error that /hi is not defined)

PR #84, opened 2020-10-19
CAS-275 Handle error when message is sent during cool down interval

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-275

### Description
Changes retry logic in a way that requests that return permanent errors from the server are not retried. As a result, user is instantly notified about message delivery errors when he/she tries to send them during cool down period. Ideally, on the UI side, there should be a timeout in `MessageInputView` preventing user from submitting a message during cool down period.

- Ensured that `cooldown` property is propagated to the UI
- Disabled retries for permanent errors including cool down interval errors (status code - 403, stream code - 60)

To be merged after this  [PR to the LLC](https://github.com/GetStream/stream-chat-android-client/pull/111) is merged and released.

### Checklist

- [X] have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [X] PR targets the `develop` branch
- [ ] Changelog updated with client-facing changes
- [X] New code is covered by unit tests
- [X] Reviewers added

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/ChannelData.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/ChannelData.kt
@@ -28,6 +28,8 @@ public data class ChannelData(var type: String, var channelId: String) {
     var deletedAt: Date? = null
     /** all the custom data provided for this channel */
     var extraData: MutableMap<String, Any> = mutableMapOf()
+    /** minimum interval between user messages (in seconds) in slow down mode **/
+    var cooldown: Int = 0
 
     /** create a ChannelData object from a Channel object */
     public constructor(c: Channel) : this(c.type, c.id) {
@@ -36,6 +38,7 @@ public data class ChannelData(var type: String, var channelId: String) {
         updatedAt = c.updatedAt
         deletedAt = c.deletedAt
         extraData = c.extraData
+        cooldown = c.cooldown
 
         createdBy = c.createdBy
     }
@@ -53,6 +56,7 @@ public data class ChannelData(var type: String, var channelId: String) {
         c.extraData = extraData
         c.lastMessageAt = messages.lastOrNull()?.let { it.createdAt ?: it.createdLocallyAt }
         c.createdBy = createdBy
+        c.cooldown = cooldown
 
         c.messages = messages
         c.members = members
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/ChannelData.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/ChannelData.kt
@@ -11,7 +11,11 @@ import java.util.Date
  * A class that only stores the channel data and not all the other channel state
  * Using this prevents code bugs and issues caused by confusing the channel data vs the full channel object
  */
-public data class ChannelData(var type: String, var channelId: String) {
+public data class ChannelData(
+    var type: String,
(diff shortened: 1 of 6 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/utils/DefaultRetryPolicy.kt
```diff
@@ -2,14 +2,15 @@ package io.getstream.chat.android.livedata.utils
...
     override fun shouldRetry(
         client: ChatClient,
         attempt: Int,
         error: ChatError
     ): Boolean {
-        return attempt < 3
+        return attempt < 3 && !error.isPermanent()
     }
 
     /**
```
tschellenbach, line 16: Perhaps we should check on the error.isPermanent flag in runAndRetry? 

what point is there in allowing customers to retry permanent errors..?

livedata/src/test/java/io/getstream/chat/android/livedata/ChatErrorTest.kt
```diff
@@ -38,4 +38,10 @@ internal class ChatErrorTest : BaseConnectedIntegrationTest() {
         val error = ChatNetworkError.create(0, "", 500, null)
         Truth.assertThat(error.isPermanent()).isFalse()
     }
+
+    @Test
+    fun `cool down period error should be permanent`() {
+        val error = ChatNetworkError.create(60, "", 403, null)
```
zsmb13, line 44: Would be a good idea to extract all these numbers to meaningful constants, to make them easier to interpret

livedata/src/main/java/io/getstream/chat/android/livedata/ChannelData.kt
```diff
@@ -28,6 +28,8 @@ public data class ChannelData(var type: String, var channelId: String) {
     var deletedAt: Date? = null
     /** all the custom data provided for this channel */
     var extraData: MutableMap<String, Any> = mutableMapOf()
+    /** minimum interval between user messages (in seconds) in slow down mode **/
+    var cooldown: Int = 0
```
JcMinarro, line 32: We should try to avoid mutable property as far as we can

livedata/src/main/java/io/getstream/chat/android/livedata/entity/ChannelEntity.kt
```diff
@@ -62,6 +62,8 @@ internal data class ChannelEntity(var type: String, var channelId: String) {
     var deletedAt: Date? = null
     /** all the custom data provided for this channel */
     var extraData = mutableMapOf<String, Any>()
+    /** minimum interval between user messages (in seconds) in slow down mode **/
+    var cooldown: Int = 0
```
JcMinarro, line 66: Same here

JcMinarro (changes requested):

livedata/src/main/java/io/getstream/chat/android/livedata/utils/DefaultRetryPolicy.kt
bychkovdmitry: Moved the check to `runAndRetry`

JcMinarro (approved):

PR #85, opened 2020-10-20
Fix/event handler

tschellenbach:
The batch event update system has become quite complex. This PR simplifies it which makes it:

- harder to make coding mistakes which forget to store users, or other required data
- easier to correctly handle events

While working on this I noticed the following issues with the existing code:

- If multiple events update a channel or messages, we sometimes ignored everything but the last update
- ReactionNewEvent and other reaction events didn't store reaction.user
- NotificationChannelDeletedEvent didn't set the user

Needs to be updated by Oleg to support the new repository structure

### DIFF
--- a/livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
+++ b/livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
@@ -3,8 +3,7 @@ package io.getstream.chat.android.livedata
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.models.User
-import io.getstream.chat.android.livedata.entity.ChannelEntity
-import io.getstream.chat.android.livedata.entity.MessageEntity
+import io.getstream.chat.android.livedata.extensions.updateLastMessage
 import io.getstream.chat.android.livedata.extensions.users
 
 /**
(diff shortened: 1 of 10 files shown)


livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    suspend fun fetch() {
+        messageMap = domainImpl.repos.messages.select(messagesToFetch.toList()).associateBy { it.id }
+        channelMap = domainImpl.repos.channels.select(channelsToFetch.toList()).associateBy { it.cid }
+        fetchCompleted = true
+    }
+
+    suspend fun execute() {
+        require(fetchCompleted) { "be sure to run batch.fetch before calling this method" }
```
samiuelson, line 111: In case`fetch()` was already called and then another `add..()` functions called the `fetchCompleted` is already set to `true` allowing to skip another `fetch()` call before calling `execute()`. 

E.g.  
```kotlin 
EventBatchUpdate(domainImpl).apply {
  fetch()
  add...()
  execute()
}
```

This `require()` assertion doesn't sound look safe.

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,119 @@
...
+ * as a second step, load the required data for batch updating using
+ * batch.fetch()
+ *
+ * third, add the required updates via
+ * addUser, addChannel and addMessage methods
+ *
+ * fourth, execute the batch using
+ * batch.execute()
+ */
+internal class EventBatchUpdate(private val domainImpl: ChatDomainImpl) {
```
samiuelson, line 29: The instructions above makes me feel that some kind of a builder pattern would be more suitable here instead. 

All the `add...()` methods could be provided by `BatchUpdate.Builder` and `fetch()` would be done in `BatchUpdate.Builder.build()`.

Eventually, this would allow the instance of `BatchUpdate` to have immutable, private properties only, making the `BatchUpdate.execute()` operation thread-safe. 

WDUT?

samiuelson (changes requested):

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
samiuelson:
```kotlin
internal class EventBatchUpdate.Builder(private val domain: ChatDomain) {
  fun add...()
  fun add...()

  fun build() {
    //fetch all the required data here and create EventBatchUpdate
    return EventBatchUpdate() // EventBatchUpdate would be an immutable class
  } 
} 
```
```kotlin
internal class EventBatchUpdate(/*only immutable props here*/) {
  fun execute()
}
```

I'm thinking about something like above.

tschellenbach: pro: type system prevents you from using it the wrong way
con: it's a bit more complex

In most languages the simple solution would be preferred. Though in the kotlin ecosystem I think the builder pattern looks more natural. Going to do that, fits more with the ecosystem.

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,119 @@
...
+        return messages[messageId] ?: messageMap[messageId]
+    }
+
+    fun addMessage(messageEntity: MessageEntity, messageUsers: List<User>) {
+        // ensure we store all users for this channel
+        addUsers(messageUsers)
+        messages[messageEntity.id] = messageEntity
+    }
+
+    fun addUsers(newUsers: List<User>) {
```
carterhudson, line 80: Maybe rename the methods that insert or manipulate the update maps to be more explicit. Something like `addUsers` -> `registerUsersForUpdate`.

carterhudson (commented):
Mostly just disambiguation comments.

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,119 @@
...
+        getCurrentChannel(cid)?.let {
+            it.updateLastMessage(MessageEntity(message))
+            addChannelEntity(it, emptyList())
+        }
+    }
+
+    fun addChannel(channel: Channel) {
+        // ensure we store all users for this channel
+        addUsers(channel.users())
+        // TODO: this overwrites members which in the case when you have > 100 members isn't the right behaviour
```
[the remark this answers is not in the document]
tschellenbach: members have pagination, which is something we ignore here. So as soon as you have more members than the pagination limit (100 by default) things get a little wonky

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,113 @@
...
+ * third, add the required updates via
+ * batch.addUser, addChannel and addMessage methods
+ *
+ * fourth, execute the batch using
+ * batch.execute()
+ */
+internal class EventBatchUpdate private constructor (private val domainImpl: ChatDomainImpl, private val channelMap: Map<String, ChannelEntity>, private val messageMap: Map<String, MessageEntity>) {
+    val users: MutableMap<String, User> = mutableMapOf()
+    val channels: MutableMap<String, ChannelEntity> = mutableMapOf()
+    val messages: MutableMap<String, MessageEntity> = mutableMapOf()
```
samiuelson, line 32: We have multiple props there having the same type, e.g.:
```kotlin
val messages: MutableMap<String, MessageEntity>
val messageMap: Map<String, MessageEntity>
```
It would be great to give them more meaningful names.

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    internal class Builder {
```
samiuelson, line 34: It would make the outer class more readable to have the inner builder class at the bottom or outside.

samiuelson (approved):

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,109 @@
...
+        fun addToFetchMessages(id: String) {
+            messagesToFetch = messagesToFetch + id
+        }
+
+        fun addUsers(usersToAdd: List<User>) {
+            users = users + usersToAdd
+        }
+
+        suspend fun build(domainImpl: ChatDomainImpl): EventBatchUpdate {
+            val messageMap: Map<String, Message> = domainImpl.repos.selectMessages(messagesToFetch.toList()).associateBy(Message::id)
```
zsmb13, line 63: We used to create this `Map` by passing in users in addition to the messages, the users from all `UserEvent`s:

```kotlin
val messageMap = domainImpl.repos.messages.select(messagesToFetch.toList(), users.toMap()).associateBy { it.id }
```

We're now adding those users to the `users` set with `addUsers ` above, but not using them anywhere.

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,109 @@
...
+ * batch.execute()
+ */
+internal class EventBatchUpdate private constructor(
+    private val domainImpl: ChatDomainImpl,
+    channelMap: Map<String, Channel>,
+    messageMap: Map<String, Message>
+) {
+    private var users: Map<String, User> = emptyMap()
+    private var channels: Map<String, Channel> = channelMap
+    private var messages: Map<String, Message> = messageMap
```
zsmb13, line 35: We might as well merge these with the ctor parameters, no need to have them separated

livedata/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    internal class Builder {
+        private var channelsToFetch = setOf<String>()
+        private var messagesToFetch = setOf<String>()
+        private var users = setOf<User>()
+
+        fun addToFetchChannels(cIds: List<String>) {
+            channelsToFetch = channelsToFetch + cIds
```
zsmb13, line 43: In this case, we're adding _a lot_ of elements to these lists, and we're definitely only doing it on one thread at a time. Keeping these collections mutable and adding to them instead of allocating new ones for each function call would be a good idea.

JcMinarro (approved):
LGTM

zsmb13 (approved):
^ We'll address the comments left over in another PR

